PR #32, opened 2014-11-26
Writable attributes

gekorob:
Modified to add the possibility to have writable attributes. 
Don't know if this agree with the validation project philosophy, but we need this capability in a plain ruby project, just to change an invalid attribute or to set an attribute after validation process.

Thanks
   Roberto

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@
 *.a
 mkmf.log
 .greenbar
+.idea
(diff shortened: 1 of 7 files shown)


jodosha:
@gekorob We have just merged #37 Which allows to define validations separately from attributes. Please have a look at https://github.com/lotus/validations#validations

The merged approach is more flexible because it just uses Ruby.

Closing this, and thank you **a lot** for bringing this useful use case.

gekorob:
Thanks to you and to the lotus community for your work. I'll try to do better next time.

Issue #35, opened 2014-12-04
Nested attributes validation


jodosha:
Ref https://github.com/lotus/controller/issues/53

PR #37, opened 2014-12-29
Refactor to allow objects to be initialised independently

stevehodgkiss:
Changes based on the discussion in https://github.com/lotus/validations/issues/36

- [x] Remove attribute tracking from the Lotus::Validations provided constructor, and provide a generic accessor setter constructor that can be overwritten easily.
- [x] Build the list of attributes and values to be validated lazily, rather than in the constructor. (`#read_attributes`).
- [x] Move type coercion functionality to setters (run separately from validations).
- [x] Fix broken tests
- [x] Extract `.validates` method
- [x] Move `.attributes` to a separate module (`AttributeDefiner`)


jodosha:
@stevehodgkiss That `type` removed in https://github.com/lotus/validations/commit/20909be9cbc27fbd07023ccdb06e3bc2c5099919 should stay there. The purpose is to test format validations in combination with coercion. I'm keen to change Utils again in order to keep that spec to pass as it was.

stevehodgkiss:
@jodosha cool, reverted... because type coercion now happens first, the attribute should always be nil or an integer, so it's really only testing that the wasn't coerced to nil.

lib/lotus/validations/attribute_definer.rb
```diff
@@ -0,0 +1,273 @@
...
+    # Define attributes and their validations together
+    module AttributeDefiner
+      def self.included(base)
+        base.extend ClassMethods
+      end
+
+      module ClassMethods
+        # Define an attribute
+        #
+        # @param name [#to_sym] the name of the attribute
```
joneslee85: This is my first time see this documenting convention `#to_sym`, does that mean we could parse in object that respond to `to_sym`?

lib/lotus/validations.rb
```diff
@@ -20,6 +22,7 @@ module Validations
     # @see http://www.ruby-doc.org/core/Module.html#method-i-included
     def self.included(base)
       base.extend ClassMethods
+      base.send(:include, AttributeDefiner)
```
jodosha:
```ruby
base.class_eval do
  extend ClassMethods
  include AttributeDefiner
end
```

;)

lib/lotus/validations/attribute_definer.rb
jodosha: @joneslee85 Yes. BTW, We have more spread around the Lotus codebase.

jodosha:
@stevehodgkiss This is :+1: for me.

I think that the usage of `Validator` in `Validations#valid?` could serve as an extension point for nested validations (#35). If we make it working **both** with **set of validations** (flatten) and **validators** (nested).

joneslee85:
@stevehodgkiss :clap: great work mate :+1:

stevehodgkiss:
@joneslee85 thanks mate!

@jodosha I'm happy for this to be merged when you're ready. There's probably a few places where the docs need updating so thanks for offering to take care of that :)

jodosha:
As future reference, benchmarks and object allocations: https://gist.github.com/jodosha/2fff07c32369ebc2c919

This implementation boosts `Validations#initialize`, and adds negligible perf penalty to `#valid?`. :+1: 

Because (custom) coercion is a shared problem between Validations and Model, we should probably think to an unified implementation in Utils. The main benefit is to have a consistent behavior for coercions.

At the beginning `Utils::Kernel` module functions seemed enough to me, but coercions are slightly different.

stevehodgkiss:
@jodosha here are my thoughts on how we could improve the coercer in utils (in the context of http params):

1. After coercion happens, verify that it actually worked. For example, if we want a DateTime, after the coercer does it's thing, check that we in fact get a DateTime. This would help Lotus Controller guarantee that the param would be either the specified type or nil. An error should occur if coercion fails.
2. If we give an array to an attribute expecting a string, we end up with essentially `[].to_s`. Similarly if we have a `Tempfile`, `.to_s` produces something like `"#<File:0x007f95b5ba6ce0>"`. I think we should raise an error instead, because in the context http params this is likely to be someone changing inputs trying to find vulnerabilities, or just a broken application.

PR #40, opened 2015-01-08
Don’t coerce to nil for blank string values

stevehodgkiss:
Controller is [testing for this behaviour](https://github.com/lotus/controller/blob/master/test/integration/full_stack_test.rb#L24), so it makes sense to ensure the correct behaviour here as well.

Rather than make the `return nil if blank_value?` line nasty with more conditionals, let's settle for less than ideal behaviour on string coercions. I don't think `[]` should successfully coerce to a string, but that should probably be dealt with in the coercer.

### DIFF
--- a/test/validations/coercions_test.rb
+++ b/test/validations/coercions_test.rb
@@ -22,4 +22,14 @@ describe Lotus::Validations::Coercions do
     result = Lotus::Validations::Coercions.coerce(Integer, '')
     result.must_equal nil
   end
+
+  it 'returns an empty string when an empty string is given to coerce to String' do
+    result = Lotus::Validations::Coercions.coerce(String, '')
+    result.must_equal ''
+  end
+
+  it 'returns nil when an array is given to coerce to a string' do
+    result = Lotus::Validations::Coercions.coerce(String, [])
+    result.must_equal nil
+  end
 end
--- a/test/validations/coercions_test.rb
+++ b/test/validations/coercions_test.rb
@@ -30,6 +30,12 @@ describe Lotus::Validations::Coercions do
 
   it 'returns nil when an array is given to coerce to a string' do
     result = Lotus::Validations::Coercions.coerce(String, [])
-    result.must_equal nil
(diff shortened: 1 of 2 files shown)


joneslee85:
:+1:

AlfonsoUceda:
:+1:

test/validations/coercions_test.rb
```diff
@@ -22,4 +22,20 @@
...
+
+  it 'returns an empty string when an empty string is given to coerce to String' do
+    result = Lotus::Validations::Coercions.coerce(String, '')
+    result.must_equal ''
+  end
+
+  it 'returns nil when an array is given to coerce to a string' do
+    result = Lotus::Validations::Coercions.coerce(String, [])
+    # This is not ideal behaviour
+    result.must_equal '[]'
```
AlfonsoUceda: [].to_s # => "[]"
{}.to_s # => "{}"

it isn't the ideal behaviour but it's the ruby default, maybe it should be like this.

stevehodgkiss: In terms of accepting untrusted input from the web, if the app receives a `[]` when it expects a string it's probably due to a malicious request. That's why I prefer the behaviour in Virtus:

```ruby
irb(main):001:0> SignUpForm.new(username: [])
Virtus::CoercionError: Failed to coerce attribute `username' from [] into String
        from /opt/rubies/2.1.5/lib/ruby/gems/2.1.0/gems/virtus-1.0.3/lib/virtus/attribute/strict.rb:19:in `coerce'
        from /opt/rubies/2.1.5/lib/ruby/gems/2.1.0/gems/virtus-1.0.3/lib/virtus/attribute/coercible.rb:14:in `set'
```

jodosha:
@stevehodgkiss Thank you, merged! :+1:

test/validations/coercions_test.rb
AlfonsoUceda: ok I understand

PR #41, opened 2015-01-08
Back attribute accessors with Utils::Attributes and support non-whitelisted attributes for Lotus::Action::Params

stevehodgkiss:
A work in progress based on the discussion here https://github.com/lotus/controller/pull/69 Initially that PR had knowledge around whitelisting/not whitelisting. This pulls that functionality into validations. Whether it actually belongs here I'm still not sure, but that's open for discussion :)

This PR changes accessors to be backed by a `Utils::Attributes` hash.

### DIFF
--- a/lib/lotus/validations/attribute_definer.rb
+++ b/lib/lotus/validations/attribute_definer.rb
@@ -1,5 +1,6 @@
 require 'set'
 require 'lotus/utils/class_attribute'
+require 'lotus/utils/attributes'
 
 module Lotus
   module Validations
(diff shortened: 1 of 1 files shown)


joneslee85:
LGTM :+1: but then I am feeling uneasy that we do it here.

jodosha:
@stevehodgkiss Thanks for taking care of this.

I'm wondering if only the changes introduced by https://github.com/stevehodgkiss/validations/commit/9a28f2ec5396b863e3358971768705cafbe895af are enough to support Params.

I'm going to experiment only with that basic change, and I'll report the results here.
In the meantime, can you please explain the rationale behind the two subsequent commits?

stevehodgkiss:
@jodosha the changes in 9a28f2e could be enough if `Params` takes care of assigning attributes in non-whitelisting mode, but then you'd still need to override `to_h` to return all values in that scenario.

dbd5b5d - is required to store unknown attributes when no attributes are specified. Otherwise they'd all be ignored, which is not the behaviour we want in Controller.
07eaddb - was in response to running the controller test suite against this validations branch. As far as I can tell, Controller expects `Params#to_h` to return a hash with stringified keys.

jodosha:
@stevehodgkiss I make it working only with the first commit, yay! :100: 

Can you please remove the other two commits and modify the first with the `||=` operator?

```ruby
      def initialize(attributes)
        @attributes ||= Utils::Attributes.new
        attributes.to_h.each do |key, value|
          writer = "#{ key }="
          public_send(writer, value) if respond_to?(writer)
        end
      end
```

jodosha:
@stevehodgkiss Merging this. Thank you!

PR #42, opened 2015-01-08
Ensure whitelisted assignment of attributes doesn't include ruby methods

stevehodgkiss:
This PR whitelists the assignment of attributes to the set of defined attributes (`public_send(writer, value) if self.class.defined_attributes.include?(key.to_s)`), and also fixes a couple of other issues that came up because of this:

- confirmation attributes weren't being added as attributes, now they are.
- defined_attributes has been changed to make inheritance and inclusion work as expected.


jodosha:
@stevehodgkiss another great catch! Thank you!

PR #43, opened 2015-01-09
Nested attributes and validations

### LINKED ISSUE #35: Nested attributes validation

stevehodgkiss:
This PR implements [nested attributes](https://github.com/lotus/validations/issues/35) and validations by building an anonymous class with lotus validations included for the nested attribute. This works because when the nested hash is assigned, the coercer instantiates the anonymous type with the hash, building a nested validator with all it's attributes.

To make the top level `params.valid?` take into account nested validators, there's a new nested validation/method in `Attribute` which uses a new `errors.set` method to assign the returned `Errors` object to the attribute. This means you can query for errors like this `validator.errors.for(:address).for(:line_one)`. What do you think to this interface for nested validation errors?

Closes #35

### DIFF
--- a/lib/lotus/validations.rb
+++ b/lib/lotus/validations.rb
@@ -203,6 +203,13 @@ module Lotus
       errors.empty?
     end
 
+    # Validates the object.
+    #
+    # @return [Errors]
+    #
+    # @since x.x.x
+    # @api private
+    # @see Lotus::Attribute#nested
     def validate
       validator = Validator.new(defined_validations, read_attributes, errors)
       validator.validate
--- a/lib/lotus/validations.rb
+++ b/lib/lotus/validations.rb
@@ -198,10 +198,14 @@ module Lotus
     #
     # @since 0.1.0
     def valid?
-      validator = Validator.new(defined_validations, read_attributes)
-      @errors = validator.validate
+      validate
 
       errors.empty?
+    end
+
+    def validate
+      validator = Validator.new(defined_validations, read_attributes)
+      @errors = validator.validate
     end
 
     # Iterates thru the defined attributes and their values
--- a/lib/lotus/validations.rb
+++ b/lib/lotus/validations.rb
@@ -204,8 +204,8 @@ module Lotus
     end
 
     def validate
-      validator = Validator.new(defined_validations, read_attributes)
-      @errors = validator.validate
+      validator = Validator.new(defined_validations, read_attributes, errors)
+      validator.validate
     end
 
     # Iterates thru the defined attributes and their values
--- a/lib/lotus/validations/attribute.rb
+++ b/lib/lotus/validations/attribute.rb
@@ -45,7 +45,18 @@ module Lotus
       # @api private
       # @since 0.2.0
       def validate
-        _run_validations
+        presence
+        acceptance
+
+        return if skip?
+
+        format
+        inclusion
+        exclusion
+        size
+        confirmation
+        nested
+
         @errors
       end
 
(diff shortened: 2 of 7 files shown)


joneslee85:
LGTM +1

jodosha:
@stevehodgkiss Thanks for this, because @joneslee85 gave his thumb up, I merged it.

stevehodgkiss:
When attempting to access a nested attributes errors *via the parent*, and there are no errors for that nested object, you have to check if there are any errors before trying to reach further down, or a no method error will occur. Let me try and explain better with an example:

```ruby
class MyValidator
  include Lotus::Validations

  attribute :book do
    attribute :title, type: String, presence: true
    attribute :author do
      attribute :name, type: String, presence: true
      attribute :favourite_colour
    end
  end

end

validator = MyValidator.new({})
validator.valid?
# => false
validator.errors.for(:book).for(:author).for(:name)
# => [#<Lotus::Validations::Error:0x007fc71a86a928 @attribute=:name, @validation=:presence, @expected=true, @actual=nil>]

validator = MyValidator.new({book: { title: "Lotus in Action", author: { name: "Luca" } }})
validator.valid?
# => true
validator.errors.for(:book)
# => []
validator.errors.for(:book).for(:author).for(:name)
# NoMethodError undefined method #for on Array
```

This means you have to be careful (check if there are any) when reaching deep into nested validations. I don't like this.

So, what do we want the interface to look like for accessing or aggregating nested validation errors?

One option could be to flatten them at each level. So for example:

```ruby
validator.errors
# => [#<Lotus::Validations::Error:0x007fbe4a0bd2a8 @actual=nil, @attribute="book.author.name", @expected=true, @validation=:presence>]
validator.book.errors
# => an Errors with keys like "author.name"
# => [#<Lotus::Validations::Error:0x007fbe4a0bd2a8 @actual=nil, @attribute="author.name", @expected=true, @validation=:presence>]
validator.book.author.errors
# => [#<Lotus::Validations::Error:0x007fbe4a0bd2a8 @actual=nil, @attribute="name", @expected=true, @validation=:presence>]
```

Thoughts?

cc @jodosha @joneslee85

jodosha:
@stevehodgkiss The original idea of Validations, when they were just one level was:

```ruby
class BookValidator
  include Lotus::Validations

  attribute :title,   presence: true
  attribute :price, presence: true, inclusion: 1..100
end

validator = BookValidator.new(title: nil, price: 2000)
validator.valid? # => false

validator.errors.for(:title) # => presence error
validator.errors.for(:price) # => inclusion error

validator.errors.each do |error|
  # yield all the flatten errors from title and price
  # ...
end
```

In the example above `validator.errors` is able to fetch errors for a single attribute via `#for`. At the same time, it was able to flatten all those errors in one level when bulk operations are executed (eg `#each`).

This is important for presentational purposes. Imagine you want to show all the violations on a page, your code can be clean like iteration above, without worrying about the nested levels.

This is the reason why `Error` responds to `#attribute`.

```erb
<% validator.errors.each do |error| %>
  <li><%= error.attribute %> expected: <%= error.expected %> is: <%= error.actual %></li>
<% end %>
```

Can we reproduce the same behavior?

jodosha:
@stevehodgkiss Another aspect that we haven't considered yet is attributes name clashing when flatten errors are returned.

```ruby
class Signup
  include Lotus::Validations

  attribute :programming_language do
  end
end

validator = Signup.new(name: nil, programming_language: { name: nil })
validator.valid?

validator.errors.each do |error|
  # ...
end
```

It the example above, the block will yield two similar errors: `error.attribute => :name` etc..

Do you think we should add another information to `Error`?

stevehodgkiss:
I've opened up a PR with a WIP to continue the discussion about nested errors https://github.com/lotus/validations/pull/46

I think that when pulling from nested validator objects, we need to build a new Error with an attribute namespaces by the parents... i.e. `programming_language.name`.

PR #46, opened 2015-01-17
Flatten nested errors

stevehodgkiss:
- Fixes an issue where by requesting errors for an attribute (with for) the Errors object would then think it had errors (with .any?)
- Flattens nested errors based on the parent attribute name.
- `#attribute` returns the nested attribute name, `#attribute_name` always returns the de-namespaced attribute name.

### DIFF
--- a/lib/lotus/validations/attribute.rb
+++ b/lib/lotus/validations/attribute.rb
@@ -200,8 +200,10 @@ module Lotus
       def nested
         _validate(__method__) do |validator|
           errors = value.validate
-          unless errors.empty?
-            @errors.set @name, errors
+          if errors.any?
+            errors.each do |error|
+              @errors.add "#{@name}.#{error.attribute}", error
+            end
           end
           true
         end
--- a/lib/lotus/validations/attribute.rb
+++ b/lib/lotus/validations/attribute.rb
@@ -202,7 +202,9 @@ module Lotus
           errors = value.validate
           if errors.any?
             errors.each do |error|
-              @errors.add "#{@name}.#{error.attribute}", error
+              namespaced_attribute_name = "#{@name}.#{error.attribute}"
+              new_error = Error.new(namespaced_attribute_name, error.validation, error.expected, error.actual)
+              @errors.add namespaced_attribute_name, new_error
             end
           end
           true
--- a/lib/lotus/validations/attribute.rb
+++ b/lib/lotus/validations/attribute.rb
@@ -200,12 +200,10 @@ module Lotus
       def nested
         _validate(__method__) do |validator|
           errors = value.validate
-          if errors.any?
-            errors.each do |error|
-              namespaced_attribute_name = "#{@name}.#{error.attribute}"
-              new_error = Error.new(namespaced_attribute_name, error.validation, error.expected, error.actual)
-              @errors.add namespaced_attribute_name, new_error
-            end
+          errors.each do |error|
+            namespaced_attribute_name = "#{@name}.#{error.attribute}"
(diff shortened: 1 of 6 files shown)


lib/lotus/validations/attribute.rb
```diff
@@ -200,8 +200,10 @@ def size
       def nested
         _validate(__method__) do |validator|
           errors = value.validate
-          unless errors.empty?
-            @errors.set @name, errors
+          if errors.any?
```
AlfonsoUceda: you can remove this if, if errors if empty the each loop won't be executed.

AlfonsoUceda:
for now :+1:

lib/lotus/validations/attribute.rb
```diff
@@ -200,8 +200,9 @@ def size
       def nested
         _validate(__method__) do |validator|
           errors = value.validate
-          unless errors.empty?
-            @errors.set @name, errors
+          errors.each do |error|
+            new_error = Error.new(error.attribute, error.validation, error.expected, error.actual, @name)
```
joneslee85: 4 params is abit hard to follow, keyed params would be great for this constructor.

lib/lotus/validations/error.rb
```diff
@@ -41,9 +58,10 @@ class Error
       #
       # @since 0.1.0
       # @api private
-      def initialize(attribute, validation, expected, actual)
-        @attribute, @validation, @expected, @actual =
-          attribute, validation, expected, actual
+      def initialize(attribute_name, validation, expected, actual, namespace = nil)
+        @attribute_name, @validation, @expected, @actual, @namespace =
```
joneslee85: how about instance variables in many lines separately? According to benchmark by Pete Yandel, 1 line assignment is much slower

joneslee85:
:+1:

lib/lotus/validations/error.rb
stevehodgkiss: Yeah I would prefer these to be on one line too. Will update...

joneslee85:
Looking great!

:+1:  from me

Waitin' for @jodosha 's review

jodosha:
I think this PR deserves a public recap of the discussion happened between me and @stevehodgkiss 

The goals are:

  * Keep bulk operations around `validator.errors.each`. It yields a flatten list of errors.
  * Nested attributes can cause name clashing (see Example 1), `Error#attribute` will return the complete name of the attribute (Example 2).
  * The "dot notation" described above is unhandy if we want only to get the attribute name. For this reason we have introduced `#attribute_name` (Example 3).

```ruby
class Signup
  include Lotus::Validations

  attribute :programming_language do
  end
end

validator = Signup.new({})
validator.valid?

# Example 1
validator.errors.each do |error|
  error.attribute # => Without this PR, it always returned 'name'. This makes hard to understand where it belongs to (top level or 'programming_language').
end

# Example 2
validator.errors.each do |error|
  error.attribute # => first iteration: "name", second iteration: "programming_language.name".
end

# Example 3
validator.errors.each do |error|
  error.attribute_name # => first iteration: "name", second iteration: "name".
end
```

We decided to add a fifth argument to `Error#initialize`, in this way, `Error` is responsible to understand and provide simple and complex names via the "dot notation".

jodosha:
The other notable change is the notation to fetch nested errors:

```ruby
# OLD
validator.errors.for(:address).for(:line_one)

# NEW
validator.errors.for('address.line_one')
```

jodosha:
This looks good :+1: Thanks @stevehodgkiss Merging.

PR #50, opened 2015-02-06
Removed support for MRI ruby versions lower than 2.2

AlfonsoUceda:
ditto

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,13 @@
 language: ruby
+sudo: false
+cache: bundler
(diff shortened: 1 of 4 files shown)


jodosha:
@AlfonsoUceda Sorry, but I have to close this PR. Because I'm already working on it
It isn't just a matter of changing the CI settings, but also to get rid of `Lotus::Utils::Attributes` and use symbols.

AlfonsoUceda:
ok perfect!

PR #55, opened 2015-03-24
Fixed lotus entity integration bug

### LINKED ISSUE #52: Compatability with Lotus::Model/Entity
Seeing consistency break down when including/using validations with `Lotus::Model`'s `Entity` mixin. The interface to a defined entity changes subtly so as to break expectations. See the following example:

```ruby
class DummyModel
  include Lotus::Entity
  attributes :name, :email
end
# => [:name, :email]

DummyModel.attributes
# => #<Set: {:id, :name, :email}>

m = DummyModel.new name: "Luke", email: "test@me.com"
# => #<DummyModel:0x007fdc8d135b48 @email="test@me.com", @name="Luke">

m.to_h
# => {:id=>nil, :name=>"Luke", :email=>"test@me.com"}
```

Vs the output of:

```ruby
class OtherModel
  include Lotus::Entity
  include Lotus::Validations
  attribute :email, presence: true
end
#=> [:name]

OtherModel.attributes
#=> #<Set: {:id, :email, :name}>
# Note, attributes includes the id here...

o = OtherModel.new name: "Luke", email: "test@me.com"
#=> #<OtherModel:0x007fc03ba9c1f0 @attributes=#<Lotus::Utils::Attributes:0x007fc03ba9c1c8 @attributes={"name"=>"Luke", "email"=>"test@me.com"}>>
# The internal structure (or at least the output of the inspect method) has changed

o.to_h
#=> {:email=>"test@me.com", :name=>"Luke"}
# What happened to the id field? `.attributes` tells me I should be able to expect it returned...
```

I can note that if I specify `attribute :id` at the same time as the validators, I can get it to return the id as well, but this is not consistent with how Lotus::Model acts.

AlfonsoUceda:
Closes #52

### DIFF
--- a/lib/lotus/validations/attribute_definer.rb
+++ b/lib/lotus/validations/attribute_definer.rb
@@ -254,6 +254,16 @@ module Lotus
         #   signup = Signup.new(password: 'short')
         #   signup.valid? # => false
         def attribute(name, options = {}, &block)
+          _attribute(name, options, &block)
+        end
+
+        # Define an attribute
+        #
+        # @see Lotus::Validations::AttributeDefiner#attribute
+        #
+        # @since x.x.x
+        # @api private
+        def _attribute(name, options = {}, &block)
           if block_given?
             define_nested_attribute(name, options, &block)
             validates(name, {})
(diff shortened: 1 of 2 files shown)


AlfonsoUceda:
@plukevdh can you test if this works for you?

jodosha:
@AlfonsoUceda It looks good to me. Thank you! :+1:

Issue #54, opened 2015-03-24
Default values for attributes?


PaBLoX-CL:
I was using `lotus/validations` in my gem, I liked the syntax sugar of attribute, with their validations. But I found out that there's no way to include default data. It seems (to me at least) something basic, so maybe I'm just missing something obvious (I read the verbose explanation on how to use only the validations, but using that I have to sacrifice almost all the sugar).

Is there already a way to do it?

AlfonsoUceda:
@PaBLoX-CL could you provide an example where the default options is useful?

PaBLoX-CL:
Sure, thought I believe I didn't explain it too well :/, hopefully it will be clearer after the example:

```ruby
class Character
  include Lotus::Validations

  attribute :name, presence: true
  attribute :race, presence: true
  attribute :age, type: Integer
  attribute :level, default: 1
  
end
> Character.new(name: 'Legolas', race: 'elf')
#<Character:0x007f5feaf6adc8 @attributes=#<Lotus::Utils::Attributes:0x007f5feaf6ada0 @attributes={"name"=>"Legolas", "race"=>"elf", "level"=>1}>>
```

I basically meant to create default values when instantiating the class.

Right now, I'm using this:

```ruby
class Character
  include Lotus::Validations

  attribute :name, presence: true
  attribute :race, presence: true
  attribute :age, type: Integer
  attribute :level
  
  def initialize chars = {}
    chars[:level] = 1 unless chars[:level]
    super(chars)
  end
end
```

AlfonsoUceda:
I think is useful provide default values, what do you think @jodosha @joneslee85

jodosha:
@PaBLoX-CL Hello and thanks for starting this discussion.
I see validations like a way to deal with user input for a specific use case, [not to build models](http://lucaguidi.com/2014/10/23/introducing-lotus-validations.html). In your example, you're clearly using validations to define a business rule.

I'm not sure if this proposal is a good idea, but I'm here to listen. Would you mind to expand a little bit your use case? How often did you had to override `#initialize`? Is that painful to override `#level` in this way?

```ruby
DEFAULT_LEVEL = 1

def level
  super || DEFAULT_LEVEL
end
```

PaBLoX-CL:
I totally understand your point and I agree with it.

[...]

PR #56, opened 2015-03-25
Added default value behaviour

AlfonsoUceda:
Added behaviour to support #54

### DIFF
--- a/lib/lotus/validations/attribute_definer.rb
+++ b/lib/lotus/validations/attribute_definer.rb
@@ -278,14 +278,12 @@ module Lotus
         # @since 0.2.2
         # @api private
         def define_attribute(name, options)
-          type = options.fetch(:type) { nil }
-
-          define_accessor(name, type)
+          define_accessor(name, options)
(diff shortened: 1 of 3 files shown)


lib/lotus/validations/attribute_definer.rb
```diff
@@ -330,9 +328,16 @@ def define_writer(name)
 
         # @since 0.2.2
         # @api private
-        def define_reader(name)
+        def define_reader(name, options = {})
           define_method(name) do
-            @attributes.get(name)
+            attribute = @attributes.get(name)
+            if attribute.nil? && !options[:type].nil? && !options[:default].nil?
```
AlfonsoUceda: I don't like a lot this implementation (the code), we have 3 cases and we support three cases, what do you think guys?
```
attribute :name, type: String, presence: true, default: 1
```
```
attribute :level, default: 1
```
```
attribute :level
```

jodosha:
Closing this. Ref https://github.com/lotus/validations/issues/54#issuecomment-90463121

PR #59, opened 2015-04-22
Refactor attribute definer for entity

dsnipe:
Since we have now separate method in Entity, that define attribute accessor, we can use it, and not override whole `attributes` function.
That can also lead to bug, because dirty tracking using new approach, while `validation` override whole `attributes`, so `define_attr_accessor` never invoked.

### DIFF
--- a/lib/lotus/validations/attribute_definer.rb
+++ b/lib/lotus/validations/attribute_definer.rb
@@ -420,21 +420,15 @@ module Lotus
           @defined_attributes.merge(attributes.map(&:to_s))
         end
 
-        # @since x.x.x
-        # @api private
+        # Override attribute accessors function.
         #
-        # @see Lotus::Model::Entity#attributes
-        def attributes(*attrs)
-          if attrs.any?
-            attrs = Lotus::Utils::Kernel.Array(attrs)
-            attrs.each do |attr|
-              _attribute(attr)
-              attr_accessor(attr) if define_attribute?(attr)
-            end
-            self.attributes.merge attrs
-          else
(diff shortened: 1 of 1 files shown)


AlfonsoUceda:
:+1:

jodosha:
@dsnipe Thanks for this PR. Merging this without a proper test coverage is dangerous, because it uses a private method from another gem. In the past we experienced regressions due to the lack of sync with Entity implementation in https://github.com/lotus/model

Can you please add an integration test of dirty tracking here?: https://github.com/lotus/validations/blob/master/test/integration/lotus_entity_compatibility_test.rb

Once this will be done, I'm :+1: for merging this. Thank you :cat2:

jodosha:
@dsnipe @AlfonsoUceda I'm going to merge it now and to write additional tests.

This fixes Lotus::Model builds https://travis-ci.org/lotus/model/builds/61315554 and https://travis-ci.org/lotus/model/builds/61523677, relative to https://github.com/lotus/model/pull/182 and https://github.com/lotus/model/pull/184 respectively.

This regression was introduced by https://github.com/lotus/validations/pull/55

jodosha:
@dsnipe @AlfonsoUceda @joneslee85 DirtyTracking doesn't work with validations. I've tried to implement it here, but it's a mess. I'm for keeping partial compat with Entity for now.

dsnipe:
@jodosha I'm busy in that week, but hopefully will find some time on the next and dig in in that problem.

jodosha:
@dsnipe That would be great. Thank you! :+1:

PR #64, opened 2015-05-18
Introduce invalid? method as an opposite to valid?

deepj:
Sometimes I need to check if a model is __invalid__ rather than __valid__. In this case I prefer this form to using `if !model.valid?` or `unless model.valid?`. I hope you consider this as an useful method.

Any comments are welcome!

BTW: I am not sure where I would have put a test for this case. So I hope I put it into a right place. Openly, I am a bit confused by the structure of tests.


joneslee85:
I am good with this proposal :+1:

AlfonsoUceda:
:+1:

PR #66, opened 2015-07-07
Fix documentation

### LINKED ISSUE #65: Documentation: Acceptance
IMHO the documentation/example of acceptance validation has a small error.

It states that _truthy_ values will make this validation pass. Empty-string ```""```is _true_, in ruby.
Since ```acceptance``` forwards to ```Lotus::Kernel::Boolean``` empty String is considered _false_.

In other words: acceptance does not evaluate according to ruby truthiness but according to Lotus::Utils::Kernel.Boolean

pascalbetz:
Closes #65

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ validations.
 
 #### Acceptance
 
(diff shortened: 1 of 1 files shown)


jodosha:
@pascalbetz Thanks for your help. Merging this! :+1:

PR #68, opened 2015-07-28
Allows attribute `type` to be specified as `Array[Person]`

krisleech:
Opening this PR to start discussion.

I would like to be able to do this:

```ruby
class Person
  attr_reader :name

  def initalize(attributes); 
    @name = attributes.fetch(:name)
  end
end

class MyForm
  include Lotus::Validations

  attribute :people, type: Array[Person]
end

form = MyForm.new(people: [{ name: 'Kris', name: 'Lindsey' }]
form.people.first.name # => 'Kris'
form.people.all? { |p| p.is_a?(Person) } # => true
```

This PR is a quick hack, there are no tests. If you think it is an okay direction - handling a coercer which is a one element Array, as a special case - then I will submit a proper PR with tests.

In theory you could also support something like `type: Array[Person, Fruit, Mineral]` if the collection was not homogeneous but can't see a usecase for this and it would probably be confusing.

### DIFF
--- a/lib/lotus/validations/coercions.rb
+++ b/lib/lotus/validations/coercions.rb
@@ -22,6 +22,8 @@ module Lotus
       def self.coerce(coercer, value, &blk)
         if ::Lotus::Utils::Kernel.respond_to?(coercer.to_s)
           ::Lotus::Utils::Kernel.__send__(coercer.to_s, value, &blk) rescue nil
+        elsif coercer.is_a?(Array)
+          value.map { |v| coerce(coercer[0], v) }
         else
           coercer.new(value, &blk)
         end


krisleech:
Discussion here: https://discuss.lotusrb.org/t/allow-lotus-validations-to-coerce-members-of-an-array/120

duncanita:
Hi, I have problem with validation of params like this
```
{ users: [ { name: 'bob'}, {name: 'Joe'}] }
```

jodosha:
@krisleech @duncanita I've looked closely at this issue and for me is a :-1:

What would that syntax mean in terms of code? Would that `Array[<Object>]` be an object that could be validated (respond to `#valid?`) or is just a type notation?

If it's the first case, how to handle the error messages with the "dot syntax" for attributes (eg `signup.user.address.city`)? If you receive `{ people: [ {name: 'Alice'}, { name: '' } ]}`, how to return the error message for the single wrong entry?

If it's just a type notation, should the validation pass if we're able to coerce it? Imagine we have a validation like this: `dates: Array[Date]`, would data like: `{ dates: ['2020-01-01'] }` be valid? Theoretically yes, but it lacks of flexibility because you may also want to add further validation rules.

I think this should be implemented via a custom validation. See #72

PR #73, opened 2015-10-06
Changed Person to PersonForm to avoid misunderstood with Entity

hlegius:
As discussed in lotus' chat, I'm proposing to change all `Person` objects in Readme to PersonForm.

I don't know if worth modify `Signup` to `SignupForm` as well. If is the case, I can change as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,8 @@ Or install it yourself as:
 
 `Lotus::Validations` is a set of lightweight validations for Ruby objects.
 
+It behaves like a Form Object pattern which decouple your Input validation from Entities or other domain objects.
+
 ### Attributes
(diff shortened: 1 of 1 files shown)


joneslee85:
cc @jodosha

jodosha:
@hlegius That `Person` isn't necessary a form object, it can be any kind of object, entity included.
Please remember that the scope of this gem is about providing validations, where **you** (as dev) you think it will work best for you. By merging this PR would limit how we communicate this and so, the scope of the lib.

In other words, I don't want to tell that `lotus-validations` it's only about params and forms, because it has a wider field.

@joneslee85 :+1:

PR #82, opened 2015-11-19
Added blank_value? to size predicate

### LINKED ISSUE #81: Empty parameter values with defined size attribute are not valid
Hello there!

I've found a bug within `size` and blank values. If I define `size` attribute and try to pass the attribute as blank/empty, it will fail.

```ruby
class User
  attribute :name, size: 8..10
end

User.new({name: '', other: 'foo'}).valid? # <= false
User.new({name: '12345678', other: 'foo'}).valid? # <= true
User.new({other: 'foo'}).valid? # <= true
```

```ruby
class User
  attribute :name
  attribute :other, size: 2..10
end

User.new({name: '', other: 'foo'}).valid? # <= true
```

How should it be, IMO:

```ruby
class User
  attribute :name, size: 3..10
  attribute :other_param, presence: true, size: 2..10
end

User.new({name: '', other_param: 'abc'}).valid? # <= true
User.new({name: 'a', other_param: 'abc'}).valid? # <= false
User.new({name: 'abc', other_param: 'abc'}).valid? # <= true
User.new({name: 'abc'}).valid? # <= false
User.new({other_param: 'abc'}).valid? # <= true
```

If you guys agrees, I'd like to fix it :+1:

hlegius:
This Pull Request added the ability to ignore blank values for `size` predicate as well occurs for other predicates i.e. presence.

The only exceptional case down to Enumerable's implementations which if defined will be treated as value as it is. i.e:

```ruby
class AwesomeValidator
  include Lotus::Validations

  attribute :name, type: String, size: 8..10
  attribute :tags, type: Array,  size: 2..10
end

AwesomeValidator.new.valid? # < true
AwesomeValidator.new(name: '12345678', tags: '').valid? # < true
AwesomeValidator.new(tags: ['a', 'b']).valid? # < true
AwesomeValidator.new(name: nil).valid? # < true

AwesomeValidator.new(name: 'ooo').valid? # < false
AwesomeValidator.new(tags: []).valid? # < false
AwesomeValidator.new(name: 'abc', tags: []).valid? # < false
```

@solnic [has argued about](https://github.com/lotus/validations/issues/80#issuecomment-158085889) Conditional Validations and 'magical' validations/coercions which may introduce more complexity into this kind of issue. Maybe it worth more discussion later /cc @jodosha 

Closes #81

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,6 +67,35 @@ person.email # => "me@example.org"
 person.age   # => raises NoMethodError because `:age` wasn't defined as attribute.
 ```
 
+#### Blank Values
+
+The framework will treat as valid any blank attributes, __without__ `presence`, for both `format` and `size` predicates.
+
+```ruby
+require 'lotus/validations'
+
+class Person
+  include Lotus::Validations
+
+  attribute :name,    type: String, size: 5..45
+  attribute :email,   type: String, size: 20..80, format: /@/
+  attribute :skills,  type: Array,  size: 1..3
+  attribute :keys,    type: Hash,   size: 1..3
+end
+
+Person.new.valid?                             # < true
+Person.new(name: '').valid?                   # < true
+Person.new(email: '@').valid?                 # < true
+Person.new(skills: '').valid?                 # < true
+Person.new(skills: ['ruby', 'lotus']).valid?  # < true
+
+Person.new(skills: []).valid?                 # < false
+Person.new(keys: {}).valid?                   # < false
+Person.new(keys: {a: :b}, skills: []).valid?  # < false
+```
+
+If you want to _disable_ this behaviour, please, refer to [presence](https://github.com/lotus/validations#presence).
+
 ### Validations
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -67,6 +67,35 @@ person.email # => "me@example.org"
...
+
+  attribute :name,    type: String, size: 5..45
+  attribute :email,   type: String, size: 20..80, format: /@/
+  attribute :skills,  type: Array,  size: 1..3
+  attribute :keys,    type: Hash,   size: 1..3
+end
+
+Person.new.valid?                             # < true
+Person.new(name: '').valid?                   # < true
+Person.new(email: '@').valid?                 # < true
```
AlfonsoUceda: is this example valid? `"@" isn't blank`

hlegius: You're right. By a mistaken, I've thought one thing and did another totally different. Good catch! Thanks :+1:

jodosha:
@hlegius Thank you, merging this. :+1:

PR #85, opened 2016-01-08
Empty String should not be accepted, according to the docs.

pascalbetz:
https://github.com/lotus/validations/issues/84

Test to demonstrate the problem + fix.


davydovanton:
:+1:

PR #88, opened 2016-01-19
Nested Validations Resolved

jac33k:
Hey there!

I've tried Lotus for one of my backend services (suits perfect, btw) and I faced the problem with nested objects' validation.

I want to keep the information about the object (like validations) in separate classes (so DSL for nesting objects was not satisfying for me). Let me show you an example:
```ruby
class Payment
  include Lotus::Validations

  attribute :recipient_account,  type: RecipientAccount
end

class RecipientAccount
  include Lotus::Validations

  attribute :iban, presence: true
  attribute :bic,  presence: true
end
```
Even if the coercion works fine for this example, there's no info about nested object validation, so I had to hack it this way:
```ruby
class Payment
  include Lotus::Validations

  attribute :recipient_account,  type: RecipientAccount

  def valid?
    [ recipient_account ].all?(&:valid?) # I have more nested attributes
    # and here goes the check for the rest of primitive values
  end
end
```
Looks ugly, huh? (and changing the framework's default behaviour (in most cases) tells you that you shouldn't have been using framework at all)

I've resolved it in my fork (more details and examples in tests and ReadMe) and now sending the PR for review. Let me know if that might be helpful for any of Lotus' users.

P.S. This PR includes nested coercion and validations for `Array[YourClass]` - it was useful for me the way it is, maybe should inside a collection wrapper (didn't need that).

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,6 +442,31 @@ error.name           # => "address.street"
 error.attribute_name # => "street"
 ```
 
+Also you can use validations nested in Lotus Validations object.
+```ruby
+class Address
+  include Lotus::Validations
+
+  attribute :street,      presence: true
+  attribute :city,        presence: true
+  attribute :country,     presence: true
+  attribute :postal_code, presence: true, format: /.../
+end
+
+class ShippingDetails
+  include Lotus::Validations
+
+  attribute :full_name, presence: true
+  attribute :address,   type: Address
+end
+
+validator = ShippingDetails.new
+validator.valid? # => false
+```
+
+Errors are handled the same way as for nested block syntax.
+```
+
 ### Composable validations
 
 Validations can be reused via composition:
--- a/README.md
+++ b/README.md
@@ -442,10 +442,10 @@ error.name           # => "address.street"
 error.attribute_name # => "street"
 ```
 
-Also you can use validations nested in Lotus Validations object.
+Also you can use validations nested in Hanami Validations object.
 ```ruby
 class Address
-  include Lotus::Validations
+  include Hanami::Validations
 
   attribute :street,      presence: true
   attribute :city,        presence: true
@@ -454,7 +454,7 @@ class Address
 end
 
 class ShippingDetails
-  include Lotus::Validations
+  include Hanami::Validations
 
   attribute :full_name, presence: true
   attribute :address,   type: Address
--- a/README.md
+++ b/README.md
@@ -465,7 +465,6 @@ validator.valid? # => false
 ```
 
 Errors are handled the same way as for nested block syntax.
-```
 
 ### Composable validations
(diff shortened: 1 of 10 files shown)


khasinski:
:+1:

patrykfischer:
:+1:

justi:
:+1:

madsheep:
:sheep:

ja999:
:beers:

joneslee85:
@jac33k can you please rebase?

jac33k:
@joneslee85 done, I've noticed slight decrease of coverage, because of one (not used) method for `NestedAttributes` class - should I be worried or was it expected?

README.md
```diff
@@ -442,6 +442,30 @@ error.name           # => "address.street"
 error.attribute_name # => "street"
 ```
 
+Also you can use validations nested in Hanami Validations object.
```
joneslee85: I would avoid using subject 'you' here if possible

lib/hanami/validations/attribute_definer.rb
```diff
@@ -294,6 +295,12 @@ def define_attribute(name, options)
           define_accessor(name, type)
           defined_attributes.add(name.to_s)
 
+          if attribute_validable?(type)
```
joneslee85: s/validable/validatable

lib/hanami/validations/coercions.rb
```diff
@@ -22,6 +22,8 @@ module Coercions
       def self.coerce(coercer, value, &blk)
         if ::Hanami::Utils::Kernel.respond_to?(coercer.to_s)
           ::Hanami::Utils::Kernel.__send__(coercer.to_s, value, &blk) rescue nil
+        elsif coercer.is_a?(Array)
+          value.map { |v| coerce(coercer[0], v) }
```
joneslee85: add a space before opening block

joneslee85:
cc @jodosha @AlfonsoUceda

lib/hanami/validations/attribute_definer.rb
jac33k: :+1:

README.md
jac33k: Any particular reason? Just curious ;)

lib/hanami/validations/coercions.rb
jac33k: not sure where should I place this space?

README.md
SleeplessByte: `Validations can be nested in Hanami::Validations objects`
^
statement

`Also you can use validations nested in Hanami Validations object`
^
action

jodosha:
Fixed by #100

PR #96, opened 2016-03-29
Custom validations

cabeza-de-termo:
Support for custom validations:
- Refactored and reified each attribute validation to run validations preserving their order of declaration
- Support for custom validations with an fully capable protocol
- Rewrote built-in validations to use the same protocol as custom validations
- Applications can add new built-in validations and compose them from existent validations

Please see the custom validations [API](https://github.com/cabeza-de-termo/validations/tree/custom-validations#custom-validations)


jodosha:
Implemented by #100 https://github.com/hanami/validations#custom-predicates

PR #97, opened 2016-03-30
Validation messages

cabeza-de-termo:
A very simple addition to display validation error messages.

It detaches the message formatting from the validations and errors responsibility and moves it to the presentation classes.

It

* allows to define and override the default validation messages
* allows to access the default validation message as `error.to_s`
* allows to override the message for a validation type only in the context of a presentation class, such as a View, Helper or, in frameworks that overlap responsibilities, a Controller
* allows to override the message for a single attribute validation only in the context of a presentation class, such as a View or Helper. This way the same validation, for instance `:presence`, can use the default message for attributes that are displayed in text fields and a different one for an attribute that is displayed in a select box.

Here you can read the [Api](https://github.com/cabeza-de-termo/validations/tree/validation_messages_poc#validation-messages)


jodosha:
@cabeza-de-termo Thanks for this PR. :smile: 

We can't merge it for several reasons:

  * We're evaluating to use `dry-validation` as backend, which already has this feature
  * There isn't support for `i18n`
  * It relies too much external components to configure the messages
  * The public API will increase a lot and has a _hard learning curve_
  * The code increases of 30% for a single feature
  * Configurations are global (eg. `Hanami::Validations::Messages.library`). This makes it impossible to use in Hanami _Container_ projects, where multiple applications can coexist together in the same Ruby process without overriding each other their configurations.

Thanks for understanding. :smile:

cabeza-de-termo:
Sure, no prob. I'll wrap it in a gem on its own.
Thanks for pointing out the issues, I'll address them in the future.
Thank you!

Issue #90, opened 2016-02-12
Format with empty/blank values does not coerce to nil


hlegius:
Hello,

As described by @wafcio in [his gist](https://gist.github.com/wafcio/164b1191d2d689bfa088), he was trying to validate a value with `format` attribute. If pass an empty string, format returns `valid true` (which is correct), but does not coerce into `nil`. If we provide `type: Integer` attribute, it will return 0, which isn't the same of nil of blank.

```ruby
class Person
  include Hanami::Validations

  attribute :number, format: /\A\d+\z/
end

person = Person.new(number: '')
p.valid? # => true
```

I was wondering if we should introduce blank/empty string to nil conversion in all occasions as well.

jodosha:
@hlegius Hello and thanks for taking care of this. Your suggestion is good and can work. If it won't take you too much time please implement it. :+1: 

In the long run, we should probably add two different layers of validations. A "HTTP" layer that solves the problem of "everything is a string", which can be used for `Action::Params` and this kind of problems reported by @wafcio.

A second layer for "core" types, where we're not trying to guess the input and coerce it, but only to verify that the given data is already of the expected type.

---

How this can solve the problem described? As now, the coercion of the types is implemented by `Hanami::Validations::Coercions`, and under the hood by `Hanami::Utils::Kernel`. The semantic of the latter is: "try to do your best to convert this input into the target type, otherwise raise a `TypeError`". But for `Validations` we suppress that error and return `nil`.

This implementation is "too smart": it tries to do the conversion, but then gives up and returns `nil`. So for an input `""` (empty string), we internally read a `nil` value. We skip validations on `nil` values and here's the source of our problem.

Now, this mechanism is in place because it tries to solve two kind of problems: HTTP and Core types in at the same time.

---

**My suggestion is to decouple them**. When an object includes `Hanami::Validations` it only gets the "core" set of coercion rules. Do we expect an `Integer`? Can be the input safely converted? If so, do it, otherwise complain. This is the logic of `Utils::Kernel`.

We should then introduce another module like `Hanami::Validations::Http`, which is included by default by `Action::Params` and it tries to handle string inputs and than make use of the core layer.

---

This can fix the root problem, when validations are used on objects, without the implications of coercible HTTP params.

hlegius:
@jodosha 

> Your suggestion is good and can work. If it won't take you too much time please implement it. :+1: 

[...]

Issue #80, opened 2015-11-08
Conditional validations


darkbaby123:
I found there's no conditional validations like `validates :name, presence, if: :some_condition`. IMO it's used very frequently. Curious the reason it's not implemented in Lotus validations. Is it already in plan? Or there's other way you recommend more to do the same thing?

joneslee85:
@darkbaby123 atm there is a PR #71 that allows users to define custom validator which can work around this issue.

I like your suggestion and open for the PR submission. The validator should take in a proc/lambda OR the symbol name of a private condition method

jodosha:
@darkbaby123 Hello and thanks for this suggestion. We won't to support conditionals for action callbacks, validators, because they increase the complexity both of the frameworks and applications.

If you use validations [around use cases](http://lucaguidi.com/2014/10/23/introducing-lotus-validations.html), and not in entities/models, you'll almost eliminate this need. Now, I don't know your specific use case, so please share with us, and we can find a solution for you of for the framework.

As @joneslee85 we'll offer soon a more flexible approach for validations, so it could be helpful for you. But my opinion on conditionals is :-1:

darkbaby123:
@jodosha I watched [your blog post](http://lucaguidi.com/2014/10/23/introducing-lotus-validations.html) before I submit this issue, and I'm not use validations in entities/model level in my recent project (all plain Ruby object + validation mixin, it's a Rails project). I admit that most of time I don't need the conditional validations, but they're still useful in some of my cases:

## Case 1

I have an API that responsible to connect some device (hardware) to current user, and I use a single API endpoint to handle all 3 connection types. It's like this:

```ruby
{
  conn_type: 'a/b/c',
  quick_code: 'xxx', // should be present in type a
  uuid: 'xxx', // should be present in type b
}
```

In this case I can use 2 different validators, but a conditional validation can make it simpler.

## Case 2

In my another project I have a "sign contract" page, the user need to choose a "sign via" radio button to select sign type, he should either enter:

1. Driver's licence and state
2. SSN

The validation in this logic is depend on what "sign via" he choose. So I did something like this:

```ruby
validates :signer_driver_lic, :signer_state, presence: true, if: :sign_via_driver_lic?
validates :signer_ssn, presence: true, if: :sign_via_ssn?
```

I think conditional validations are not used frequently (if so then it's wrong way to use validations), but in some cases it make sense. A simple "if" is better than an extra abstraction (e.g. a different validator object).

[...]

Issue #79, opened 2015-11-05
PORO validator


magicalbanana:
This would be a feature request but would be nice if we could implement this to be just a PORO validator as well.

Perhaps do something like Virtus where you can set the constructor or the mass assignments.

jodosha:
@magicalbanana Hey, can you please tell us more about your idea? Thank you!

magicalbanana:
@jodosha take a look at [Virtus](https://github.com/magicalbanana/veto) it's a good PORO validator really and I like the idea that the validation is handled outside or is of a seperate concern.

Because (correct me if I am wrong) but the idea of Lotus is to make it modular right? So the validations should not create the attributes but rather just validate the attribute when passed an entity. Because I think right now it's like an Entity with a built-in validator right?

[Virtus](https://github.com/solnic/virtus) has some configs you can pass when including the module.

(Cherry-picking extensions)[https://github.com/solnic/virtus#cherry-picking-extensions]

If we had something like this for the validator it would be kewl because it would make our code readable. We would have attributes defined and state the validations we're going to be using.

I hope this makes sense. I can help making this a feature but would like to have a code walk prior to starting (it's just easier for me to get the run down of the code from the original coder).

deepj:
And what `Lotus::Validations` replace by [Dry::Validation](https://github.com/dryrb/dry-validation)?

I see here a good fit with philosophy of Lotus. And it can be used more widely such as forms, params and so on...

magicalbanana:
Also, I've am in the process of making mine as well [PoroValidator](https://github.com/magicalbanana/poro_validator)

If you guys could give feedback that would be great. It's still WIP but the core classes are already there, I am just now making the validations.

jodosha:
@deepj Starting from next week, I'll explore Dry::Validation for that purpose :+1:

solnic:
Thanks for suggesting dry-validation. We'll be looking into how it can help Lotus. One thing I should mention is that dry-validation is for validating *data*, for now we only support objects responding to `[]` but next version will support validating objects through attr readers, too. Personally I don't think it's gonna be a common thing to do, but since it is simple to support, we're gonna add that.

jodosha:
@solnic Thank you :+1:

jodosha:
Closed by #100

PR #98, opened 2016-04-20
Predicates

### LINKED ISSUE #72: Custom validation plugin API
Firstly I would like to say that I *love* Lotus validations. I have actually almost entirely replaced my ActiveModel validations with it, and it has been a dream to work with.

Since I use Lotus separately from the rest of the Lotus framework, there is one feature I really would love to have: the ability to create custom validations. Looking in the source, there are a finite number of accepted validation types in the API, specifically:

```ruby
      VALIDATIONS = [
        :presence,
        :acceptance,
        :format,
        :inclusion,
        :exclusion,
        :confirmation,
        :size,
        :type,
        :nested
      ].freeze
```

I would like to be able to create my own validations, and append to this list. For example, say I wanted to validate the authenticity of a token, it would nice to be able to create my own validator for this:

```ruby
attribute :jwt_token, authenticity: true
```

This would allow me to create my own logic to verify the authenticity of a token, and have it validated by Lotus.

Currently, to achieve this, I can do some somewhat nasty monkey-patching by overriding `#valid?`

```ruby
def valid?
  super()
  validate_jwt_token_authenticity
  errors.empty?
end

def validate_jwt_token_authenticity
  if external_jwt_validation
    errors.add :jwt_token, Lotus::Validations::Error.new(:jwt_token, :presence, true, false)
  end
end
```

Would you guys accept a PR which created a way in which users could add custom validations to the validation set? I would like to add this functionality if possible :smile: 

Thoughts? Concerns? Would love to hear what you guys think about this!

- Ian

jodosha:
## Current State

The current API that uses options to define validations suffers from **rigidity**, **poor extensibility**. Developers don't have control on the execution order of the validations (_presence_, then _size_), which caused problems with blank values. Nor they can decide to skip a validation during certain conditions.

## dry-rb

Talking with dry-rb Community, I’ve discovered a new and powerful approach.

It works with input hashes and lets us to define a set of validation rules **for each** key/value pair. These rules are wrapped by lambdas that can only return `true` or `false` to state that a specific key is valid or not. To do that, we translate business requirements into predicates that are chained together with Ruby _faux boolean logic_ operators (eg. `&` instead of `&&`).

Think of a signup form. We need to ensure data integrity for the  `name` field with the following rules. It has to be: present **and** a string **and** its size must be greater than 3 chars, but lesser than 64. Here’s the code, **read it aloud** and notice how it perfectly expresses our needs for `name`.

```ruby
require 'dry-validation'

schema = Dry::Validation.Schema do
  validates(:name) { present? & str? & size?(3..64) }
end
result = schema.call(name: "Luca")
result.success? # => true
```

There is more than that, [`dry-validation`](http://dry-rb.org/gems/dry-validation/) puts emphasis on _type safety_ as a fundamental approach for a reliable process.

In the past months, I've experimented with `dry-validation` as backend for our library because it solves most of the problems described above. While `dry-v` is a powerful tool, I decided to take the good parts of it and to not adopt it.

### API Stability

There isn’t a way to wrap `dry-v` API, that means **to expose its public API as our public API**. We’re going to shift our approach from _option based_ to _predicate based_. That is a major breaking change for our Community. If we can’t wrap `dry-v`, **all its breaking changes will become our breaking changes**.

One of the main Hanami values is maintenance. We aim to protect developers from low level changes by offering a stable API. For instance `hanami-controller` wraps `rack`.

### API Control

There is more to consider: if we can’t wrap, we can’t control what to expose. Each method name, each method signature, each feature wouldn’t be decided by us.

There are things that I prefer to not adopt, for instance the code for _predicate logic_ is represented in memory with a custom AST, which makes impossible to debug it.

### The Support Problem

`dry-v` is still young and there are still problems to address. This is fine, but a bug in `dry-v` will become our bug without having the chance to fix it.

### Too Many Transitive Dependencies

`dry-v` would introduce 10 transitive dependency gems. Hanami v0.8 will have 20 gems in total for the **whole framework** and brand new projects will have 30 gems, if you consider `capybara`, `rake`, `bundler` et co.

Adding 10 gems only for validations is too much.

## Custom Solution

After that long experiment with `dry-v`, I decided to try with a custom solution heavily inspired by it. This _pull request_ is the result of this new direction.

[**For the API details, please have a look at the updated README.**](https://github.com/hanami/validations/blob/predicates/README.md)

## The Forking Road

Once I announced this decision in chat, there was an argument with the dry-rb Community. According to them, We should use only one validations library in Ruby Community, otherwise it will sink.

It’s true that there isn’t unity outside of the Rails world.
We have low level frameworks that are duplicated across the _Alt Ruby Community_. We have `hanami-view`, dry has `dry-view` and Trailblazer has Cells.

At the end of the day, we should choose what’s best for us as maintainers and for our developers.

For instance, we’ll probably adopt ROM as backend for `hanami-model`, because ROM is so powerful that we can bear with this maintenance burden. For a validation library it isn’t worth the trouble.

## A Community Discussion

I’m reporting all the story here because for openness sake. **I want to hear your voice now**: what’s your technical opinion about this change? What do you think about the _predicate syntax_? Shall we merge this or adopt dry-v?

---

Closes #72 #79 #80 #90 

/cc @hanami/core-team @hanami/contributors


lib/hanami/validations/context.rb
```diff
@@ -0,0 +1,189 @@
...
+
+      def type?(expected)
+        _type?(expected)
+      end
+
+      def val(key)
+        @input.value(key)
+      end
+
+      def value
```
vyper: I think only one method... "merge" #val and #value

```ruby
def value(key = @key)
  @input.value(key)
end
```

So, we can use:

```
validates(:password) { eql?(value(:password_confirmation)) }
validates(:password) { value == "123" }
```

vyper:
I like your approach. (:

I don't like for excessive dependencies in my project, this main point in my opinion.

README.md
```diff
@@ -34,544 +34,542 @@ gem 'hanami-validations'
...
+```shell
+$ gem install hanami-validations
+```
 
 ## Usage
 
-`Hanami::Validations` is a set of lightweight validations for Ruby objects.
-
-### Attributes
+`Hanami::Validations` is a mixin that, once included by an object, adds lightweight set of validations to it.
```
cllns: "adds lightweight set" => "add __a__ lightweight set"

README.md
```diff
@@ -34,544 +34,542 @@ gem 'hanami-validations'
...
 
-The framework allows you to define attributes for each object.
+It works with input hashes and lets us to define a set of validation rules **for each** key/value pair. These rules are wrapped by lambdas that can only return `true` or `false` to state that a specific key is valid or not. To do that, we translate business requirements into predicates that are chained together with Ruby _boolean logic_ operators (eg. `&&`).
```
cllns: "lets us to define" => "lets us define"

README.md
```diff
@@ -34,544 +34,542 @@ gem 'hanami-validations'
...
-person.age   # => raises NoMethodError because `:age` wasn't defined as attribute.
+result = Signup.new(name: "Luca").validate
+result.success? # => true
 ```
 
-#### Blank Values
+There is more that `Hanami::Validations` can do: **type safety**, **composition**, **complex data structures**, **built-in and custom predicates**.
 
-The framework will treat as valid any blank attributes, __without__ `presence`, for both `format` and `size` predicates.
+But before to dive into advanced topics, we need to understand the basics of _boolean logic_.
```
cllns: "before to dive" => "before diving"

README.md
```diff
@@ -34,544 +34,542 @@ gem 'hanami-validations'
...
+```
 
-### Coercions
+NOTE: This works because `Integer` responds to `#even?`.
 
-If a Ruby class is passed to the `:type` option, the given value is coerced, accordingly.
+#### Any?
 
-#### Standard coercions
+This predicate is like `#all?`: it verifies one or more conditions (inside a block) on the elements of an array. But to make it to pass, **at least one element** must satisfy the conditions.
```
cllns: "must satisfy the conditions" => "must satisfy the condition"

README.md
```diff
@@ -34,544 +34,542 @@ gem 'hanami-validations'
...
-**Validations are triggered when you invoke `#valid?`.**
+```ruby
+require 'uri'
+HTTP_FORMAT = URI.regexp(%w(http https))
 
-#### Acceptance
+validates(:url) { format?(HTTP_FORMAT) }
+```
+
+#### Greater Than
```
cllns: I prefer `greater_than` over `gt`, since it's much more clear.

README.md
```diff
@@ -34,544 +34,542 @@ gem 'hanami-validations'
...
+| `true`        | ✔︎     |
+| `nil`         | ✗     |
+| `""`          | ✗     |
+| `"0"`         | ✗     |
+| `"1"`         | ✔︎     |
+| other strings | ✗     |
+| `0`           | ✗     |
+| `1`           | ✔︎     |
+| other numbers | ✗     |
+| other objects | ✔︎     |
```
cllns: GitHub renders these two quite differently. Perhaps you could use a different [Unicode checkmark](https://en.wikipedia.org/wiki/Check_mark)?
screen shot 2016-04-20 at 09 44 11

cllns:
Great work @jodosha! Very flexible 👌

jxxcarlson:
I'm not sure if this is the right place to comment, but I very much favor the approach suggested here.  For me, the main points are (1) the proposed new API is both elegant and powerful.  We will all enjoy using it; (2) API stability and keeping breaking changes to a _bare minimum_ is extremely important.  Development is hard enough work any way without having to go back and fix your code because some one upstream changed something. I think the dry-rb community should be very happy that their code is the source and inspiration for the proposed validation API. To conclude, I strongly favor this proposal.

apotonick:
If I've learned one thing from a decade of OSS, then it's *"Don't even think about rewriting everything yourself!"*.

I don't think it's a good idea to replicate libraries.

You underestimate the amount of work involved in maintaining your own validation library. Where the dry-rb guys can focus on exactly that, you are busy with a hundred other things to orchestrate and code. I'm in the exact same position and more than excited to use everything other projects provide (to make a real point, that's why I integrated `hanami-validations` in Reform).

Counting the number of gem dependencies is a poor metric and something I would not even consider. It's more about this:

* Do you trust the team providing the dependency?
* Is the dependency implementing what you want?
* Are you willing to communicate with the dependency team?

One of the reasons I'm losing interest in this project is because you're replicating too much. You could easily use Cells instead of `dry-view`, we provide the exact same thing with 10 years of maturity. This will end up being another rewrite of everything, and I would love to see Hanami going into the other direction, where dry-rb, Trailblazer and Hanami work for the same goal **and code**.

jxxcarlson:
Those are indeed good points

avdi:
The dry-v approach seems awfully clever. And by clever, I mean tricky. 

What is `str?`? Where does it come from?  Can I use predicate methods defined on the value itself? How?  What is the the value of self inside the validation block? How do I add new predicates?

Obviously, as a reasonably clever programmer I can find out the answers to all of these questions. My point is, it's not abundantly obvious. More importantly, it's not clear to me what is the value of these mysteries. 

Whats wrong with good  'ole sending messages to objects?

joneslee85:
I am in favour of this proposal though I find the predicate syntax a bit magical.

For example:

```ruby
validates(:name) { true }
```

With the above example, what is the value of the subject `name` inside the predicate block?

Besides, on the decision of not wrapping dry-v or adopting it, I am also with you on this for I think 1.0 is close and mitigate breaking changes should be place on top for sake of adoption.

timriley:
I'd imagine that @solnic (as dry-validation author) would have some considered responses to all your comments, @avdi, but I know he's currently on holiday with no laptop and GitHub notifications turned off. I thought it might be worth stating this here for the record :)

avdi:
Thanks for the heads up, @timriley! 

FWIW, @solnic and I used to work in the same consultancy. I know him, and I have no doubt he has very good reasons for his choices :-)

I'm just putting in a vote for those choice possibly not being right for every project. Or, perhaps he will convince me otherwise. :-)

timriley:
In @solnic's absence, I’ll make a best effort at explaining a little about dry-validation. I know this is slightly off topic here, but @avdi asked :)

I know a lot of the work that’s gone into dry-validation has been about making it:

1. Completely explicit and precise
2. Stateless
3. Flexible enough to model any sort of validation logic
4. Run nicely as a standalone unit
5. Structured such that the DSL/high-level API (which may usually entail some sort of compromise in the name of ergonomics) is decoupled from the underlying validation processing logic, which can remain powerful enough to satisfy points 1-4 above and allow the gem to be used in unforeseen ways (perhaps even in the context of hanami)

Together, I think these make for a very powerful validation library. We’re already using it successfully in multiple apps, and it’s helped us nicely tease apart what have historically been really tangled-up concerns.

It's in order to satisfy all of these requirements that it's been implemented in the way you note as tricky. Despite this, the project's still been garnering a number of new contributors doing helpful, meaningful work.

Anyway, to answer your questions specifically:

> What is `str?`? Where does it come from?

It comes from [this file in dry-logic](https://github.com/dry-rb/dry-logic/blob/master/lib/dry/logic/predicates.rb#L79-L81), which is what powers the predicate logic in dry-validation.

All predicates are completely standalone and stateless, which makes them easy to reuse (and allows you to work them specifically on your validation schemas, which you can change for each particular use case).

> Can I use predicate methods defined on the value itself? How? How do I add new predicates?

You actually [define predicate methods on the schema itself](http://dry-rb.org/gems/dry-validation/custom-predicates/), or put them in a container for reuse across schemas. This keeps the predicates connected to the place where the need to be run (the validation schemas), and not attached to the data that's being validated (because the predicates may not be appropriate for all validation contexts for that data).

> What is the the value of self inside the validation block?

AFAIK `self` inside the schema definition block is actually not all that important or useful. The definition is intended to be completely abstract, keeping the user focused on the predicate logic itself.

In this way, `self` inside the `key` / `optional` blocks is just the definition environment for specifying the predicates that should be applied to the values of those keys.

> How do I compare the value of attribute foo to attribute bar in a single validation?

You can do this with a “[high-level rule](http://dry-rb.org/gems/dry-validation/high-level-rules/)”. These rules are run in a second pass after the initial per-key validations have run. This ensures that the high-level rules run in a known-good environment (e.g. if a per-key validation requires a value to be present, we don’t want to crash any high-level rules that can rightly assume that value to be around). This approach also allows for type coercions to run prior to the high-level rules, which is great for form schemas, allowing the high-level rules to operate e.g. on proper date/time objects rather than plain strings from the form post.

Anyway, a value comparison rule would look like this:

```ruby
schema = Dry::Validation.Schema do
  key(:first).required
  key(:second).required

  rule(first_and_second_match: [:first, :second]) do |first, second|
    second.eql?(first)
  end
end

# Valid
schema.(first: "hi", second: "hi")
# => #<Dry::Validation::Result output={:first=>"hi", :second=>"hi"} messages={}>

# Invalid
schema.(first: "hi", second: "hello")
# => #<Dry::Validation::Result output={:first=>"hi", :second=>"hello"} messages={:second=>["must be equal to hi"]}>
```

If you have any thoughts about how these ideas could be more clearly expressed in our documentation, @avdi, I'd love to hear them :)

Now, as for the main issue behind this PR, as a user of dry-validation and a contributor to dry-rb, I'd certainly love to see what this powerful validation library used as widely as possible. However, I understand the hanami/dry-v development timelines aren't matching perfectly right now, and that Hanami as a "one-stop" framework wants to offer certain API stability promises.

That said, I know from watching dry-validation develop over these last few months, that building a robust validation library is a ton of work. And given validation is a really critical application boundary concern, it would be fantastic to see some of the "alt-ruby" community congregate around a shared effort (especially if it is something standalone and easily re-usable), allowing that community to have more time to build the other parts of the Ruby web app ecosystem. 

Now, if this is something that cannot happen in the near term here in hanami, I'd still hope it's something we can revisit again before too long 😄

pascalbetz:
> If I've learned one thing from a decade of OSS, then it's "Don't even think about rewriting everything yourself!".

Agree with @apotonick on this one. I've been absent in Hanami developments the last few weeks (a bit more than one month left in my current job and need to finish some projects - still interested in the ideas of Hanami!)

During that time I've been thinking about the future of Hanami a bit myself.
I don't think it is a good idea to replicate a library just to have control over it. @solnic and the ```dry-v```-team have put a lot of work into the code and by rolling our own we lose a lot of that.
Especially when it is "heavily inspired" without making considerable changes.

The resources of the Hanami team are limited. Why not profit from the hard work of others? 
I understand that for a framework it is undesired to lose control over API/Bug-fixes. But given the alternative of rolling everything our own I'd opt for integrating dry-v.

jodosha:
@apotonick 

> You underestimate the amount of work involved in maintaining your own validation library.
We did until now and nobody said a word until we opened this discussion. Again, I'm not afraid of this.

---

> Counting the number of gem dependencies is a poor metric
I absolutely trust `dry-rb` folks. The problem with transitive dependencies is simple: they can grow far away from your trust circle. What if `concurrent-ruby` folks decide they need yet another dependency and include a gem with dozen of other dependencies?

Of course this problem can happen with our existing low level deps: `tilt`, `rack`, `sequel`, `http_router` and here we're following your principle of "Don't even think about rewriting everything yourself!". Which is good, I never considered to write SQL adapters. Well, in these specific cases, we're willing to take the risk of dependencies growth, but not (sorry) for a validation library.

---

> One of the reasons I'm losing interest in this project

No offence at all. But your sentence makes me to think about the poor cohesion of _alt ruby community_.

> but I don't like the philosophy of a closed environment, where Hanami controls all gems. This is exactly why Rails is stuck

Mr @solnic mentioned this problem too. May you please elaborate a bit? Why do yo think it's a closed env? What's wrong with controlling the gems? It means for me to provide a safety net for our developers, where we guarantee that nothing changes by accident.

jodosha:
@avdi Thanks for joining this conversation. I didn't get your point: each new library has a learning curve, the goal of an author is to make that curve soft as much as possible. Both hanami-v and dry-v have good documentation where you can learn what's `str?` about. Also, the proposed changes of this PR are going in the [dry-v direction](https://github.com/hanami/validations/blob/predicates/README.md). 

---

The bottom line of this discussion is: because we think that _predicates logic_ is better than _option based logic_ (like activemodel), shall we roll our own or use dry-v behind as an engine?

Is your scope of discussion more broader than this question? If so, what's your opinion about _predicates logic_ and the proposed implementation? Thank you very much.

apotonick:
> Of course this problem can happen with our existing low level deps: tilt, rack, sequel, http_router and here we're following your principle of "Don't even think about rewriting everything yourself!".

Sorry to disagree again, but except for Sequel, that's all gems with incredibly simple APIs (Tilt has 2 public methods, and Rack, well...). A validation library has many different public classes and methods and a massive API - it will be *a lot of effort* for you guys to "keep up with dry-v", in case you're going the fork road.

And, again, sorry for my frank words, but here's my position: The alt-ruby community (I didn't even know this word exists) is still evolving. We all work on partly overlapping stuff, but every "big" project does cover its own problems. 

I was really excited about Hanami as a replacement for Rails that does *not* enforce any hard dependencies, so that it works nicely with Trailblazer and Cells (because that's my personal agenda). From what I can see, there is no attempt at all in Hanami to incorporate existing solutions. I find this sad, and also dangerous, as the Hanami team - and you know that I love every single person in this team - will never be able to achieve a comparable quality when "copying" gems. The "copied" gem will always be a few steps ahead.

When Rails accidentially "copied" Reform, I had a similar discussion, where everyone argued that it's "so much easier to re-implement existing gems and keep control", and that's why Rails is stuck in its 1.0.0 architecture whereas many other projects have moved on.

And that's my point! I don't want that to happen to Hanami. As @pascalbetz points out, we lose more than we gain from copying - we all should try to integrate our stuff wherever possible. Focus on Hanami as a kick-ass web framework that gives us the infrastructure, integrate the dry-rb libraries as this will take work load of you, and make it easy for the Trailblazers to run our stuff on top of Hanami.

Much love! :heart:

jodosha:
@joneslee85 

> With the above example, what is the value of the subject `name` inside the predicate block?

It's mentioned in the [README](https://github.com/hanami/validations/blob/predicates/README.md#current-value):

```ruby
validates(:name) { value == "Luca" }
```

> I am in favour of this proposal though I find the predicate syntax a bit magical.

What's magical specifically? That block is an anonymous Proc where the context of execution respond to certain methods. We call these methods _predicates_. For convenience's sake we provide built-in predicates to solve common use cases like `present?` or `confirmed?`. When these methods are not enough, the execution context can be expanded with [custom predicates](https://github.com/hanami/validations/blob/predicates/README.md#custom-predicates).

> Besides, on the decision of not wrapping dry-v or adopting it, I am also with you on this for I think 1.0 is close and mitigate breaking changes should be place on top for sake of adoption.

Ok. So do you think that this proposal should be improved and than it may be merged?

solnic:
Thanks for starting the discussion. I'm gonna elaborate once I'm back from holidays. There's a lot to say here.

pascalbetz:
@solnic so much about being on holidays with no computer:-)

@jodosha I totally see the problem of having dependencies and dependencies of dependencies. Losing control over the API, waiting for a urgent bug-fix, change of directions/philosophy. All valid problems. And probably everybody has been burned by a rotten dependency:-(

Some thoughts:

* For almost every aspect in my code there is smarter people out there. Even if I copy their code my code is just a dumbed down version and will not profit from these smart people in the future
* Resources are limited. Building a web framework is a gigantic effort (whom am I telling this. You did it yourself). I doubt that the benefits of tight control outweigh the costs of building your own.
* If the library has a simple interface and is not completely woven into the framework then you could replace it when really needed. Especially if the API is the one you would mimic


Besides the "buy or make" discussion: i like the predicate style but have not used it beyond "hello world"-level.

avdi:
>each new library has a learning curve, the goal of an author is to make that curve soft as much as possible. Both hanami-v and dry-v have good documentation where you can learn what's str? about. 

You can say that about anything. You can say that about all of Rails. 

One of the things I like about what I've seen in Hanami  so far is that it is so dedicated to making things obviously objects, in a way that people can instantly recruit their entire understanding of how objects work in order to build on what they see. This syntax would seem to depart from this set of values. (note: it may be no *worse* in this sense than the existing validation system, but it seems like replacing the system would be an opportunity for progress...) 

I am absolutely not going to get into a lengthy discussion  of "... Actually, if you read the documentation, you'll see that..." because my point is that when I look at a Hanami controller, nobody needs to say that to me. I can immediately call on my understanding of Ruby objects and classes to see how I might build on what is there. 

And, as you say, this may all be out of scope. For this and the above reason,  I'm going to unsubscribe from this thread. Thanks for your time!

jodosha:
@avdi Oh, now I see what you mean. Thanks for the clarification and your time as well. 😄

avdi:
Like I said, I don't want to clutter up this discussion with out-of-scope notes. But I did have some quick thoughts on designing validation systems, so for anyone who is interested I wrote them up over here: http://devblog.avdi.org/2016/04/21/some-quick-thoughts-on-input-validation/

jodosha:
@avdi 
> Like I said, I don't want to clutter up this discussion with out-of-scope notes.

I hope that my previous comment didn't make you feel unwanted or unwelcome in this thread. _Au contraire_, your opinions matter for me. You should feel welcome as anyone in this Community is. 😄

avdi:
@jodosha no, not at all. I'm speaking from my own best intentions here: I don't really know what's in scope, and I don't want to derail the conversation. :-)

jodosha:
Quoting @avdi's article:

> Here’s a hard truth of usability: you don’t get to decide what “makes sense”. In the context of Ruby, client programmers will always try to write “regular old Ruby code” inside “special” contexts.

There is one difference between h-v (this PR) and dry-v: it's about the context of execution of validation block.

#### hanami-validations

`self` is `Hanami::Validations::Context`. Each new `signup.validate` invocation gets a new instance, so the validator itself doesn't hold a state. The context **It's able to execute ANY "regular old Ruby code"**: you can replace `{ str? }` with `{ value.is_a?(String) }` and it still works.

That means you can entirely translate this _idiomatic_ code:

```ruby
validates(:name) { present? && str? && size?(3..64) }
```

with

```ruby
validates(:name) do
  !value.nil? && !value.to_s.match(//\A[[:space:]]*\z//) &&
  value.is_a?(String) &&
  value.bytesize >= 3 && value.bytesize < 64
end
```

Now, the code above is verbose, that's why we added some syntactic sugar: the predicates.

The reasons why we have `str?` (and others) are:

  1. Shorter, nicer
  2. I returns a specific error: `#<Hanami::Validations::Error:0x007f8f6aa2ddb0 @key=:name, @predicate=:type?, @expected=String, @actual=1>`
  3. If you run use the `is_a?` version it returns a generic error: `#<Hanami::Validations::Error:0x007ff06430d1d8 @key=:name, @predicate=:base, @expected=nil, @actual=false>`

#### dry-validation

`self` is an AST and **there isn't a chance to run "regular old Ruby code"**

```ruby
required(:name) { false }
```

It raises a:

```
/Users/luca/.gem/ruby/2.3.0/bundler/gems/dry-validation-2f2ab25b871d/lib/dry/validation/schema/dsl.rb:95:in `define': undefined method `to_ast' for false:FalseClass (NoMethodError)
```

@timriley @solnic there is a chance to remove the AST?

solnic:
## dry-v vs hanami-v

I've become frustrated when I saw Luca ditching dry-v because I hoped to see collaboration between our communities so that solving this difficult problem will be simpler. Seeing a clone of dry-v is, for me, a failure in collaboration. I actually don't want to see only one validation gem in ruby, diversity is important, however in this case it is clear that we want to solve this problem in the same way. This is confusing and frustrating for me as I prioritized my work on dry-v to make it robust and stable quickly so that trb and hanami can benefit.

I would also be careful when estimating the effort here. Validation and coercion is more tricky than it seems to be.

## Dependencies

Dry gems that are common like container or equalizer are so simple that dependency hell is almost impossible. Dependency on kleisli can be dropped in dry-types and concurrent-ruby is probably what you already depend on.

I really don't see this as a problem and if you do end up using rom you will pull these deps anyway...

## API concerns

Dry-v is a front-end for dry-logic, it is a DSL, literally (as in, not a couple of class macros with a ton of options). We compile rule ast into dry-logic rule object graph. Additional macros have been added to reduce boilerplate.

I decided to use predicate logic and composable rules because it's simpler, faster and deterministic. We have plenty of validation gems which work like AM::V and I don't think it makes sense to copy the same flawed design.

> @avdi  Whats wrong with good 'ole sending messages to objects?

I prefer functional design where objects can be composed like functions. It's superior to anything else I tried so far. Achieving the same goal in a more OO way would require decorating each value to provide predicate methods, furthermore return objects would have to expose predicate logic operators rather than simple true or false. With the way it works now we generate rule object graph once, it can be cached and it simply operates on the input in a functional way without creating any intermediate objects.

> @jodosha there is a chance to remove the AST?

No, but we could allow executing blocks and simply pass in a value. It would be against general philosophy of composing validations from stateless predicates and enable people to clutter rule definition with arbitrary code, so I would prefer to have that as a plugin. Not to mention that you have no idea which rule failed when arbitrary ruby code can be executed, which would make generating error messages difficult.

flash-gordon:
I completely agree with @solnic. I think AST is a more general approach and having block-syntax is just a special case of it. I can even suggest a syntax right off the top of my head:
```ruby
required(:name).assert { name.size > 1 }
```
And just as you cannot tweak assertion failure messages more than saying the-rule-foo-failed, you cannot tweak it in dry-v. Of course it is possible to go further and build a complex custom environment for evaluating assertion block but I don't think it makes much sense.

solnic:
> One of the main Hanami values is maintenance. We aim to protect developers from low level changes by offering a stable API. For instance hanami-controller wraps rack.

Hanami is a young project too. It is unrealistic to think you will provide a stable api faster than another lib that was in development for many months.

> dry-v is still young and there are still problems to address. This is fine, but a bug in dry-v will become our bug without having the chance to fix it.

A reported bug in dry-v gets fixed and we push a new release. Not sure how that's different from having "your own" validation lib. Even better, you help us on fixing stuff but you have less things to fix since you are reusing exisiting solution which is not your primary concern. win-win

> There are things that I prefer to not adopt, for instance the code for predicate logic is represented in memory with a custom AST, which makes impossible to debug it.

AST is compiled to rule object graph. These are simple callable objects that have no side-effects. Not sure how this is impossible to debug. I had to debug it plenty of times and it never ever was an issue.

I sense that you simply don't like the ast approach. I get that as it's not common. I had the same reaction when I was introduced to this via morpher project. What you're missing here is the nice benefits that come with this approach. In example we can use predicates exposed by schema object since rule compilation happens when we already have schema instance available. Rules can be easily inherited by simply setting the ast in inherited configuration, so ie a json schema can be extended by a form schema, coercion logic will remain form-specific while rules will be inherited. The abstract representation is way simpler to work with than a huge and complex object graph. Creating a rule ast from some input and asking rule compiler to give you back rule object graph is simpler than creating that same object graph manually. Especially when it's deeply nested. This is just a chaotic list of things from the top of my head.

Some random things you probably missed when looking into dry-v:

- structure validation with required and optional keys which works using exactly the same rule system. This is a really useful and powerful feature
- Adding new predicates is as simple as defining a method in schema or using a shared predicate container
- Predicate logic is extendible (yes, we can have more operators). We support conjuction, disjunction, exlusive disjunction and implication. Plain ruby does not allow you to do that so your solution is limited to ||, && and ^
- We have high-level rules that are executed only if their dependencies met requirements
- We support macros that can collapse common predicate composition into a single method call. I'll make it possible to add new macros easily as part of public API
- Error message compilation supports custom compilers, we can easily generate all kinds of error representations
- Built in support for i18n and plain yaml messages
- Integration with dry-types which encapsulate coercion logic
- Integration with dry-types which can also encapsulate domain constraints so that you will not have to duplicate that in your system
- Coercions infered from type expactations are form- or json-specific, handled by dry-types. This allows us to reduce complexity that comes with coercion logic to bare minimum and make it faster than a general purpose coercion logic (which I found to be an antipattern)
- Last but not least, dry-v follows our fp principles. It works in stateless mode, has no side-effects, lots of predicates are pure functions (although caching their results would be risky since the input is not known in most of the cases)

Hanami solution is gonna miss a lot of that. You will be spending a significant amount of time fixing various edge cases and figuring out how things should work. I wish you all the best even though building your own solution is a lose-lose situation in the short term, and an unknown in the long term.

Related to this discussion I posted a bunch of questions on the hanami forum that are my primary concerns these days when it comes to hanami project: https://discuss.hanamirb.org/t/what-is-hanami-where-is-it-going/222

jodosha:
@solnic 

> I've become frustrated when I saw Luca ditching dry-v because I hoped to see collaboration between our communities so that solving this difficult problem will be simpler.

I'm sorry for your frustration, but we're not "ditching", we're discussing. I could've have merged this work **without** having this Community discussion. But I'm here to listen to anyone and take this decision weighting everyone's opinions.

**What I cannot create, I do not understand.** (Richard Feynman)

I was stuck with dry-v integration, and tried to build the same thing with a similar API, but different implementation. This experience made me to discover small and yet important details to understand and to bring on the discussion table (like in the case of the AST). For me, was better to grasp these details before to make the change of using `dry-v`.

jodosha:
That being said. We're have two alternative solutions and it seems you folks in this discussion are pending for dry-v adoption.

Please have a look at the [branch](https://github.com/hanami/validations/tree/predicates-with-new-backend) that implements predicates with dry-v and tell me what you think.

**If there is someone against dry-v adoption, please speak now.**

dan-klasson:
I'm for not rewriting the wheel. I think `dry-validation` is an excellent gem and well maintained. As stated earlier, well maintained gems are not a burden.

cllns:
@jodosha You showed in the [`README`](https://github.com/hanami/validations/blob/predicates/README.md) what the API would look like for predicates without dry-v.

What would the API look like for `hanami-validations` backed by `dry-validation`? Basically the same as `dry-validation` except using `validates` instead of `key`?

jodosha:
@cllns Nope, that would look **exactly** the same. We can wrap top level DSL, but not nested blocks.

For instance, we can make this to work:

```ruby
class Signup
  include Hanami::Validations

  validates(:name) { ... }
end
```

But not this:

```ruby
class Signup
  include Hanami::Validations

  group(:address) do
    validates(:name) { ... } # This would raise an exception
  end
end
```

That last `validates` would raise a `NoMethodError` because we don't have control on nested blocks. For consistency sake we need to keep top-level methods the same of inner ones. So instead of `.validates`, we need `.required` etc..

AMHOL:
@jodosha I think this is just a missing interface, I can see how extending the DSL could be useful, not sure what @solnic thinks about this?

mrbongiolo:
@jodosha what's the common use case for validations in Hanami? I've checked the getting started guide, and it seems they are used internally by the `params { ... }` in an Action, am I wrong?
Is there any other place that I would add validations when using Hanami?

What about instead of writing:
```ruby
    params do
      param :name,             presence: true
      param :email,            presence: true, format: /@/, confirmation: true
      param :password,         presence: true,              confirmation: true
      param :terms_of_service, acceptance: true
      param :avatar,           size: 0..(MEGABYTE * 3)
      param :age,              type: Integer, size: 18..99
    end
```

you would have:
```ruby
    params do
      required(:name).filled
      required(:email).filled(format?: /@/, :confirmation?)
      required(:password).filled(:confirmation?)
      required(:terms_of_service).filled(:bool?, eql?: true)
      optional(:avatar).value(size?: 0..(MEGABYTE * 3))
      optional(:age).value(:int?, size?: 18..99)
    end
```

I'm not sure if that's the only case, but through the basic docs I couldn't find any other usage of `hanami-validations`. To be honest it seems that the Hanami doesn't really it's own lib for validations, this could be accomplished by `dry-v`...

Maybe you want `h-v` to do other stuff also, but then it would be a bit out of scope.

joevandyk:
This is a random vote in favor of using dry-v.

sandelius:
My two cents are that we should use as few dependencies as possible. My vote is no.

hlegius:
This thread remember me from the old days with Merb and Rails, where the first get _merged_ into the second to _avoid reinvent the wheel_. I agree with @apotonick when he said about join development forces to build something really good, but what I saw here is nobody saying: ok, I leave my own gem project behind to join you. Or maybe, let's build something new together with as a new project name or even I like to deal with layer X so, you can use me to build layer Y. What I see, though is three good projects building the same old web stack from scratch based on your own desires and techniques. One based on less dependencies, another more Functional and the last in favor of DSLs.

Personally, I am a 100% agree with @avdi thoughts about "what to expect from a framework". How close from Ruby notation is better. I mean, Pure Object-Oriented approach, objects interaction using explicit dependencies. Less magical, more explicitness.

I liked the ability to write code within validation blocks like @jodosha explained above because too much predicate-like would force me to check documentation uncountable times to learn something strictly useful for "that" lib or framework. (RSpec I am looking at you). Now and then its necessary, but sometimes it would be simple as plain Ruby notation.

If all three projects had the same vision on what is good for webdev, probably a collaboration between them would be more natural. I don't see three different projects (solving similar problems) as a problem in this case, though. So, I vote to keep hanami-v independent from dry-v.


<3

solnic:
> This thread remember me from the old days with Merb and Rails, where the first get merged into the second to avoid reinvent the wheel

Merb was consumed and killed by Rails. Nothing was merged into anything. This situation is completely different, we're discussing reusing an existing lib.

> What I see, though is three good projects building the same old web stack from scratch

We're not building the same old web stack under dry-web. As [I already mentioned in the forum](https://discuss.hanamirb.org/t/what-is-hanami-where-is-it-going/222/3?u=solnic) we're working on a stack with pluggable components. We just started with Roda but we'll support other routers, view layers etc. too.

> I mean, Pure Object-Oriented approach, objects interaction using explicit dependencies.

This is the actual problem that I see - it might be the case that our philosophies/approach differ too much and any collaboration will only end up with misunderstandings and frustration.

> Less magical

Well, every DSL is "magical" in some way. If you want to avoid magic and use objects explicitly all the time - good luck, but you're gonna end up with lots of boilerplate, but I'm sure you know that.

> more explicitness.

Actually, dry-v has extremely explicit behavior. Literally every single concept in this library (and dry-logic) is represented by an object. These objects do interact with each other. Every predicate is an object, every rule is an object, every rule composition is an object etc. Apart from that, it's *very* consistent, there are no "special cases" where we tweak some behavior based on some option at run-time, it's predicate-logic-based rule execution all the time. The only thing which adds "hidden" behavior are macros, but it's the price we have to pay if we want to reduce boilerplate (and it's worth it).

> I liked the ability to write code within validation blocks

Arbitrary blocks allowed in DSL is a recipe for messy class definitions. I've seen this so many times that I simply don't want to allow that because it wouldn't be an improvement for me. Besides, "define your own predicate method" is as simple as it can be.

> too much predicate-like would force me to check documentation uncountable times to learn something strictly useful for "that" lib or framework.

The predicates is the interface of the library, I don't see why learning it would be a problem.

> If all three projects had the same vision on what is good for webdev, probably a collaboration between them would be more natural. I don't see three different projects (solving similar problems) as a problem in this case, though. So, I vote to keep hanami-v independent from dry-v.

Two very similar validation libraries *is* a problem. It's a node-js-kind-of-a-problem with libs like "foo" and "foo-better" and "foo-better-and-faster" and "foo-better-and-more-faster-without-left-pad-dep". It's especially a problem when people involved want to solve the same problem in the same way.

solnic:
> I think this is just a missing interface, I can see how extending the DSL could be useful, not sure what @solnic thinks about this?

I'm not convinced that hiding the API in this case makes much sense so I don't think having a public API for extending the DSL with custom methods is worth the effort. dry-v is a high-level abstraction with its DSL, it has lower-level APIs too but I'm not sure if it makes sense to use it as a "backend". dry-logic is the backend.

The more I think about this the more I see that there's no real need for a hanami-specific validation library. It's obviously not my call at all but I'm just wondering what is the value of an additional DSL in hanami actions and especially what is the value of having a dedicated validation gem. People should be able to use whatever they prefer. Use Reform, dry-v or even ActiveModel::V, it should not matter from hanami pov. Hanami actions are simple objects. You could have it as simple as `params { |input| # call whatever you want }` to define params-processing.

Let me repeat - maintaining a validation library *will be a big effort* and it *will increase the development cost* and *it will slow down* hanami evolution.

This *really* boils down to this: are you confident that you're gonna build something better and you can afford the cost, or not? If the answer is "yes", then I genuinely wish you good luck and I'll be more than happy to say "hey, hanami-v turned out to be a better solution, we're discontinuing dry-v project" once you succeed.

From the very start I wanted to help because I support every effort that aims to improve the ruby ecosystem. I thought "hey, if we could provide a great validation lib for hanami they'll have a huge problem solved and can save a lot of time". Something that I did not know, is that hanami wants to wrap even such high-level APIs as validation to guard users from breaking changes. In some cases - I get that, in this case - I do not. Once dry-v hits 1.0 it'll be as reliable as hanami-v would be.

There's one more thing I should mention - Luca's concern is to avoid upgrade-nightmares with hanami like we've all experienced with rails and 3rd party gems. There's one thing that you might be missing - with rails and rails-like gems there's an insane amount of inheritance, most functionality you mix into your objects or inherit from some base class *or do both*. Gems typically don't provide standalone re-usable objects, they give you inheritance, mix in who-knows-what-kind-of-behaviors and eventually your entire system relies on inherited behaviors from 3rd party gems with very-little-up-to-zero encapsulation on your side. And THIS is the reason why upgrading apps is so painful, hanami wrapping all kinds of APIs and doing the additional work to "protect you" from breaking changes is a questionable approach. Not only does it need heaps of additional complexity, but also, at the end of the day, it still makes you couple your code to the framework. The way I see it is this: provide re-usable *objects* with *trivial* interfaces. Inheritance is *evil* in majority of the cases. The fact that a DSL in some gem changes is not a big problem! The actual problem is that when this gem extends *your objects* so that you end up with tight-coupling between your objects and 3rd-party APIs that provide *unpredictable* behaviors.

This is actually related to the "top of the tool chain" argument I asked about [here](https://discuss.hanamirb.org/t/what-is-hanami-where-is-it-going/222?u=solnic).

hlegius:
I've tried ROM and Rodakase once. Unit testing was _complicated_, because I always need to use that _global_ _container_ instance/definition. I can't isolated stuff and test each unit. Entities as dummy DTO and well, maybe sounds grumpy (not my intention) but _.call_ for everything. Really!? (Functional style, I know).

Maybe it's me and _my OO_ screaming when try functional in Ruby lang; maybe it's just personal preference. Or maybe I am dumb enough and did a wrong use of it. One thing is true: it was unpleasant. It's not about you or your skills/experience as developer @solnic, but you know that.

Leaving the Functional and Object-Oriented approach for another thread. What's the solution then?

There are three great communities here: dry, trailblazer and hanami. All people involved has experience, knowledge and even better: want to build a really good alternative for webdev in Ruby. What's missing?

Three projects. Both three wants to deliver a complete web env for code. How collaboration would work this way? Let's suppose here that Hanami uses ROM and dry-data/validation: Hanami will be almost a _fork_ from those projects with some additional flavor. For me it's different of Linux from scratch and any Linux distro. Will left for Hanami to build a view/template lib built based on dry-* gems.

Btw, I don't see h-v public interface similarity as a problem. It's like a pattern: RSpec has let dsl and does Ministest. Same public interface, another way to implement and collaborate within client code. In the other hand, if you're talking about copycat, I might agree with you.

DSL is not a problem. DSL-Driven is. You define a class and the next 15~25 lines are for DSL configuration/definition. It is a problem. It doesn't matter if those lines came from one or 10 frameworks, it will be lots of lines for dsl config.

solnic:
> I've tried ROM and Rodakase once. Unit testing was complicated, because I always need to use that global container instance/definition.

These are still very early days of dry-web (ex-rodakase), so it's difficult to try it out on your own. I don't know what you did exactly so it's hard to say much about it, but it's an OT so maybe let's chat about this in a different place. With container+auto_inject creating objects in tests is trivial, you just do `Foo.new` and deps will be injected for "integration" tests, and `Foo.new(some_mock)` when you're doing isolated tests. I'm not sure how/why you found that to be difficult, but since we're missing a testing infrastructure that provides some stuff OOTB I can see how people can fail at using these APIs. FWIW it's been working for us remarkably well.

> Entities as dummy DTO and well, maybe sounds grumpy (not my intention) but .call for everything. Really!? (Functional style, I know).

No, not `.call` for everything, but for many things - yes. There are objects that have wider interfaces, like relations or repositories (in rom). But lots of things can simply expose `call`, this is extremely simple, makes objects compose very well, and its functional nature (lack of "data-as-state", passing values to methods, avoiding side-effects) is superior to classical OO approach that I've seen and tried in the past.

We're getting super OT right now, so please ping me in another place if you'd like to learn more. This is a huge topic, and a reason why I want to write a whole book about it, so there's that.

solnic:
uhm, I missed @jodosha comment, so re this:

> I'm sorry for your frustration, but we're not "ditching", we're discussing

Now we are, I just had the wrong impression that this won't happen, so sorry about that. Besides, I'm just in a shitty mental shape, not really having much energy for OSS, so I'm probably overreacting all the time.

> I was stuck with dry-v integration, and tried to build the same thing with a similar API, but different implementation

I'm not surprised you were stuck since you tried to wrap its API when the low-level parts are not as polished as the DSL, and wrapping DSL would just be...weird :) As I mentioned before, I don't think wrapping dry-v makes any sense and I really wonder if you need hanami-v at all, just let people use whatever validation lib they want, including not using any lib and handling params with a custom, ad-hoc solution (which may happen). Having said that, I understand this might be against hanami philosophy of a full-stack framework that makes many important "decisions" for you, if validation must be included in the list of these decisions, then I suppose a hanami-v project makes perfect sense, but it could be as simple as exposing dry-v inside `params` block and provide facilities to configure dry-v while hiding its configuration APIs, so it'd be a simple facade on top of dry-v. In addition to that, you could integrate dry-v with hanami-model to provide things like validations that need a database (common case: uniqueness validation).

Anyhow, I'm pretty sure people are already tired of seeing me writing even more comments, so unless I see any other questions/concerns, I'm gonna stop now. Please consider everything we talked about, if you decide to go with dry-v, you can be damn sure you'll be relying on a well maintained library, and bugs will be fixed, any potential breaking changes will be communicated early prior releases, and we'd probably get our communities closer together and something good should come out of that, too. We can even go as far as letting you embed our docs inside your website, so that people don't even need to know that they are using dry-v. I'm open to all kinds of ideas how we can make it work best for both parties.

pascalbetz:
I'm, not sure if the discussion should be dry-v vs. H:V or even AM-V.

pascalbetz:
I'm not sure the discussion should be dry-v vs. H-V or even AM-V. Or if running ruby code inside a predicate block is good or bad. 
To me this should more be about the way that Hanami is going and the value it stands for and making a realistic decision about what amount of work our small community can perform.

And so far I like the way Hanami is going. Trying to build a strong alternative web framework that values a clean architecture. Nevertheless I think we should not try to re-invent the wheel or mimic a library just to have it fully under control (which might be an illusion anyway). If this should grow bigger then Hanami needs to either find way more commiters or be willing to integrate existing libraries that get the job done.
As an example: I don't understand we we re-write validations (yup, there might be smarter ways or limitations) while there are (to me) important features missing (I18n, anyone? I can not use Hanami for client jobs until it has proper I18n support).

So yes. I am all for re-using existing code. Even if it does only fit 95%. Either we live with 95% or help to improve the library.

IMHO only after this decision has been made by Hanami core we can make one step further and have a vote for a specific implementation (dry-v or whatever is out there doing this job, while dry-v seems to be the favorite).

jxxcarlson:
Let me just say clean architecture is a value to be valued and preserved.

I have had such a pleasant experience building now two apps with
Hanami, and it is that clarity of structure which draws me to it.
I would imagine that that is true of others also.

I don’t have the technical expertise to have any kind of opinion
on the issues being discussed, but those are my two cents worth.

One other general comment — is it not the case that with Hanami
one can use various other components?  If so, then perhaps
being conservative about what is added is a good way to go.

vasspilka:
Hello, sorry for coming late to the discussion.

I believe we all love creating, and we especially like having some spare time at the end of the day.

I also think that no software developer, community or company for that matter has the time and resources to create everything around their infrastructure. That being said it would make sense that Hanami joined forces with the dry community for some goals.

Hanami has been forming a very unique character. Its codebase is full of comments, it is using modules to include and extend methods. These allow a newcoming developer to be able to get a full understanding of the tools he is using and gives him the power to be able to contribute to the extension of the framework.

Dry does not follow these exact conventions, the goal however is for developers and communities to work together. Isn't the reason why Hanami is so readable, to enable more people to improve the framework?
Although I'm not following the dry community closely I have the feeling it creates good and quality code despite not being structurally the same as Hanami.

Now regarding the dependencies I can see that the extra libraries can a burden, a solution would be to work together and try to replace some in dry-v itself even though I can understand that not being feasible.

I am pro in using dry-v as long as we try to cooperate from both sides of the table. And be willing to give in so both communities can benefit. Hanami is a new framework and hopefully has a long way ahead, If it proves that dry-v was not well suited we can always build a validation library from scratch later. So as long as the dry community takes in account the needs of the Hanami community an vice-versa, it seems silly for both the communities to work on the same thing, on different places.

To conclude, while I am not nearly as close in technical knowledge as the rest of the people here. In the "people working on open projects" mindset it makes absolute sense to include suited dry gems the Hanami community.

vasspilka:
I can understand both @jodosha wanting to have a seamless environment for all hanami components (maybe that is the right decision) and @solnic feeling frustrated by seeing the community divide while being in a state that it can greatly benefit from working together.

We all have our needs, but lets not forget that everything we build is to bring people and information closer not the other way around.

jodosha:
@vasspilka 
> Isn't the reason why Hanami is so readable, to enable more people to improve the framework?

Make contributions easy is a side effect.

There are two main reasons. First one is API documentation: when a developer wants to know how to use a method, they can read docs or source code and get the job done.

Second reason is bug diagnosis: when someone reports a defect, we check the stack trace. Annotating private methods with `@since` tag, helps to understand better the circumstances where that problem occurred.

jodosha:
The TL;DR of this discussion is: we want to adopt `dry-validation`.
I'm closing this in favor of WIP #100

apotonick:
Great, I think this is the right decision, Luca. :heart:

cabeza-de-termo:
Hi @jodosha,

It seems like I'm a bit late :(, but here are my 2 cents anyway.

I don't think boolean predicates are a good model for validations (in particular, for form validations) for the following reasons:

##### Non trivial boolean expressions are more difficult to understand for us (programmers) than sequentials statements (at least they are for me)

Just put a &&, a || and a ! in a simple boolean expression and it becomes a logical mess:

```ruby
validates(:field) { predicate_1 && (!predicate_2 || predicate_3) }
```

That could be rewritten as

```ruby
predicate :not_predicate_2_or_predicate_3 do
	!predicate_2 || predicate_3
end

validates(:field) { predicate_1 && not_predicate_2_or_predicate_3 }
```

but now `not_predicate_2_or_predicate_3` is an artifitial artifact from the predicates point of view

##### The outcome of a validation is not a boolean value (true|false). Is a collection of preconditions that were not met by the input.

If you are driving without using the seat bealt and a broken light, you will still get 2 tickets, not just the first one ;).
Introducing both `&&`, `||` and `&`, `|` operatos in the boolean expression will make the expression even harder to understand and mantain.

Now, that being said, I like much more your approach (https://github.com/hanami/validations/blob/predicates/README.md) to boolean predicates for several reasons:

* Hanami is in control of its interface, just like you said
* Hanami may include or drop in the future different, custom features than dry-v, which can go from trivial differences like using `#less_or_equal_than?` instead of `#lteq?` to some new awesome feature we can not think about right now but we may in the future, or to optimize and simplify the interface for particular use cases we may find repeated in Hanami applications, etc
* I don't think that having a single library to fit all possible use cases (be it to validate data or for any other feature) is a good thing, but quite the opposite. That kind of libraries usually end up in bloated, hard to configure, debug and understand libraries, and that is one of the many things that made me walk away from Rails and love Hanami. Rather than trying to figure out every possible scenario in advance, the clean, expressive and well factorized design of Hanami gives me the building blocks to build my custom solution for my particular needs, and that is probably the most valuable thing about Hanami
* For those who prefer dry-v, thanks to Hanami architecture it will still be possible for them to integrate it into Hanami apps instead of Hanami::Validations
* I really liked you predicates DSL much better over dry-v's. I find it much more simple, readable and easy to grasp
* I don't really care about dependencies or the size of the code, just the APIs

jodosha:
@cabeza-de-termo Thanks fot your input. 😄 

> For those who prefer dry-v, thanks to Hanami architecture it will still be possible for them to integrate it into Hanami apps instead of Hanami::Validations

Unfortunately this isn't the case. At this point the current stable version is incomplete (bugs and technical limitations), and at the same time we have an OSS library to rely on (dry-v), which already provides what we need.

This discussion is so important because it will change the way we validate data with Hanami projects. We'll support dry-v syntax and, of course, leave the old syntax behind.

This was the right time for the switch, because we can still influence dry-v via feedbacks **and** because being us under 1.0, we're still in time for breaking changes.

jodosha:
Folks, I understand that if we leave this thread open, people will chime in in 6 months voting yes or not. We all decided that this was the right choice for Hanami, DRY and Ruby in general.

**Please use your voice, use your opinion, use your design skills to provide feedbacks over #100. Thank you.** 💚

PR #107, opened 2016-07-08
Concat I18n load-paths, otherwise they are being overridden

solnic:
I honestly don't know how this was passing on travis, but I18n is full of surprises so...:)


jodosha:
@solnic Thank you very much!

PR #111, opened 2016-08-04
Fix example for messages_path method [SKIP CI]

artofhuman:


lib/hanami/validations.rb
```diff
@@ -204,7 +204,7 @@ def messages(type)
       #   class Signup
       #     include Hanami::Validations
       #
-      #     messages :i18n
+      #     messages_path 'config/messages.yml'
```
cllns: Do we need to specify `    messages :yaml`  if we do `messages_path`?

artofhuman: I think no
https://github.com/hanami/validations/blob/master/test/messages_test.rb#L6-L9

solnic: If you provide a path then it'll be used to load custom messages that will be merged with the builtin ones thus it shouldn't be required. If it is, I'd say it's probably a bug.

jodosha: @cllns @artofhuman @solnic It isn't required. `:yaml` is the default (as in `dry-validation`).

PR #124, opened 2017-03-06
Update copyright to 2017 [ci skip]

nickgnd:


cllns:
thanks! (phew, done!)

sakuro:
FYI, in [Copyright notice](https://en.wikipedia.org/wiki/Copyright_notice#Form_of_notice_for_visually_perceptible_copies), all you have to contain in it are:

- Circle C
- The name of copyright owner
- *The first* year of publication

So, if you omit "2017", you will not need to update the notice every year.

cllns:
Oh, that's great! Thanks @sakuro. We also have the years in the README but I think we can get rid of that.

PR #127, opened 2017-03-29
Migrate from minitest to rspec

artofhuman:


spec/support/assertions.rb
```diff
@@ -0,0 +1,19 @@
...
+    module Assertions
+      def assert_successful(result, key = :foo)
+        expect(result).to be_success
+        expect(result.messages.fetch(key, [])).to be_empty
+      end
+
+      def refute_successful(result, messages, key = :foo)
+        expect(result).not_to be_success
+        expect(messages).to eq(result.messages.fetch(key))
+      end
```
katafrakt: Just an idea. Since RSpec is more about expectations than assertions, maybe it would be better to call those methods `expect_successful` and `expect_not_successful`?

artofhuman: I think it's a good idea

artofhuman:
@jodosha can we merge this?

jodosha:
@artofhuman Thank you very much! 👍

PR #128, opened 2017-04-16
Update documentation example for Form validation

davydovanton:
/cc @hanami/core

### DIFF
--- a/lib/hanami/validations/form.rb
+++ b/lib/hanami/validations/form.rb
@@ -16,8 +16,8 @@ module Hanami
     #     include Hanami::Validations::Form
     #
     #     validations do
-    #       required(:name).filled(:foo?)
-    #       optional(:location).maybe(:str?)
+    #       required(:name).filled(:str?)
+    #       optional(:location).filled(:str?)
     #     end
     #   end
     #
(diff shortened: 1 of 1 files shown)


lib/hanami/validations/form.rb
```diff
@@ -1,4 +1,4 @@
-require 'hanami/validations'
+  require 'hanami/validations'
```
artofhuman: I think it's should not be moved

davydovanton:
@hanami/core could you check folks?

lib/hanami/validations/form.rb
```diff
@@ -14,7 +14,18 @@ module Validations
     #
     #   class Signup
     #     include Hanami::Validations::Form
+    #
+    #     validations do
+    #       required(:name).filled(:foo?)
```
TiteiKo: `#foo?` is defined nowhere, I'd replace it with `#str?`

lib/hanami/validations/form.rb
```diff
@@ -14,7 +14,18 @@ module Validations
...
+    #       optional(:location).maybe(:str?)
```
TiteiKo: Because of the `#maybe` here, the following would pass... Is it worth mentionning?
```ruby
assert Signup.new(name: 'Luca', location: nil).validate.success?
```

lib/hanami/validations/form.rb
```diff
@@ -14,7 +14,18 @@ module Validations
...
+    #     end
     #   end
+    #
+    #   result = Signup.new('location' => 'Rome').validate
+    #   result.success? # => false
+    #
+    #   result = Signup.new('name' => 'Luca').validate
```
TiteiKo: I think using symbols as keys works too (needs checking). If so, it might be good to put an example with each key format.

jodosha:
@davydovanton ping 😉

davydovanton:
done :)

davydovanton:
thanks for the review! ❤️

PR #139, opened 2018-01-04
Fix typo

DSIW:


oana-sipos:
Thank you, @DSIW for fixing this!

PR #158, opened 2018-11-29
Bump dry-validation version to ~> 1.2

WojciechKo:
Hey guys!
What do you think about following the breaking change made in [dry-validation v1.2.0](https://github.com/dry-rb/dry-validation/blob/master/CHANGELOG.md#v0120-2018-05-31) and rename `Hanami::Validations::Form` into `Hanami::Validations::Params`?


WojciechKo:
I see it's related with https://github.com/hanami/validations/pull/148. @jodosha are we somehow closer to the upgrade?
Maybe we want to keep the interface the without changes but still make the upgrade of `dry-transaction`?

WojciechKo:
To speed up the merge I've dropped following the rename change of `dry-transaction` because it requires update in multiple hanami's gems.

WojciechKo:
Can't happen before the update to the newest rom.rb https://github.com/hanami/hanami/pull/973#issuecomment-443146903

PR #171, opened 2019-05-08
Upgrade to dry-validation 0.13

ianks:
This PR bumps the dependency on dry-validation to 0.13. A few gems in the Hanami ecosystem rely on this, and we were locked with 0.11, and oldcer versions of other dry gems. Hopefully this sets us up for an easier migration since dry-validation 1.0 coming out soon.


jodosha:
I'm sorry but dependencies are locked on a purpose. `dry-validation` is moving towards 1.0 with breaking changes here and there. Because `hanami-validations` is above 1.0, no breaking changes are allowed.

That means we can't upgrade to the latest `dry-validation` until `hanami-validations` will reach 2.0. Thanks for understanding.

PR #180, opened 2019-07-03
Hanami::Validator

### LINKED ISSUE #134: Make validations a callable object
At his talk at railsclub(https://www.youtube.com/watch?v=SRQVhHzW-Eo) @jodosha talked about 
 making changes to `Hanami::Validations` to make it a callable object.

I am happy to start working on this, would just like some guidance so I can move in the right direction.

Would we have to make this backwards compatible still allowing for params to be set on initialization?

I would also like to work on moving it from being a `Mixin` to being an "inheritable" class as also mentioned in the talk, but that would be a separate Issue/PR.

### LINKED ISSUE #173: Upgrade to dry-validation v1.0 or remove dependency from Hanami
I spend a lot of time today trying to upgrade hanami-validations to use dry-validation v1.0.0.rc3 with little success. It is possible to upgrade the dry-rb parts, but the predicate framework that Hanami created has an unclear upgrade path that I can tell.

This dependency is preventing me from upgrading ROM in my project, and I'm not using the models.

Could hanami-validations be made an optional dependency in Hanami? Is there work being done to prep hanami-validations for dry-validation v1.0.0?

Thanks.

jodosha:
## Overview

This PR introduces a few, important changes:

  * `Hanami::Validator` is the new main object for this gem.
  * `Hanami::Validations` is still kept for backward compatibility with `hanami` 1.x
  * This gem is compatible with latest gems from DRY/ROM
  * This gem will work both with `hanami` 1.x and 2.x
  * This gem ships with the new validation syntax and rules from `dry-validation` 1.x.
  * The versioning of this gem will NOT be tight to `hanami`, but it will follow its own SemVer. This is to keep the pace with DRY/ROM

ℹ️ Existing Hanami 1.x apps can decide to upgrade to `hanami-validations` 2.x, in order to use the latest ROM/DRY gems. Developers that will take this path MUST embrace the new validation syntax.

ℹ️ This gem is now a kind of alias of `dry-validation`. We eliminated the overlap. 🎉 

## `Hanami::Validator`

This is the new main object for this gem. It's a superclass to inherit from.
It has **exactly** the same usage of `Dry::Validation::Contract`.

```ruby
# frozen_string_literal: true

require "hanami/validations"

class SignupValidator < Hanami::Validator
  schema do
    required(:email).value(:string)
    required(:age).value(:integer)
  end

  rule(:age) do
    key.failure("must be greater than 18") if value < 18
  end
end

validator = SignupValidator.new

result = validator.call(email: "user@hanamirb.test", age: 37)
result.success? # => true

result = validator.call(email: "user@hanamirb.test", age: "foo")
result.success? # => false
result.errors.to_h # => {:age=>["must be an integer"]}

result = validator.call(email: "user@hanamirb.test", age: 17)
puts result.success? # => false
puts result.errors.to_h # => {:age=>["must be greater than 18"]}
```

For other features/usage, please refer to the updated README of this PR, or to [`dry-validation` documentation](http://dry-rb.org/gems/dry-validation).

## `Hanami::Validations`

`Hanami::Validations` is still kept for backward compatibility with `hanami` 1.x, but will be likely removed in 3.x.

## Breaking changes

  * Drop support for Ruby: MRI 2.3, JRuby 9.1.
  * New validation syntax
  * Removed custom predicates (`Hanami::Validations.predicate`)
  * Removed global custom predicates (`Hanami::Validations::Predicates`)
  * Removed messages path setting (`Hanami::Validations.messages_path=`)
  * Removed messages namespace setting (`Hanami::Validations.namespace`)
  * Removed messages engine setting (`Hanami::Validations.messages`)

---

Closes #134
Closes #173


lib/hanami/validations/form.rb
```diff
@@ -12,57 +10,65 @@ module Validations
...
     #
-    #   result = Signup.new(name: 'Luca', location: 'Rome').validate
+    #   result = Signup.new(name: "Luca", location: "Rome").validate
     #   result.success? # => true
     module Form
+      # @since 2.0.0
+      # @api private
+      class BaseValidator < Dry::Validation::Contract
+        params do
+          optional(:_csrf_token).filled(:str?)
```
AlfonsoUceda: is it possible to make this attribute custom or do we need it with this name?

solnic: This should be `optional(:_csrf_token).filled(:string)` - it will work the same, but it'll include the type spec.

jodosha: @AlfonsoUceda This is for backward compatibility with Hanami 1.x. See https://github.com/hanami/controller/blob/master/lib/hanami/action/params.rb#L119

jodosha: @solnic I changed it with: https://github.com/hanami/validations/pull/180/commits/02596c30962ac63a7ce67da15207ba650f6e8b25

lib/hanami/validations.rb
```diff
@@ -76,13 +45,13 @@ module ClassMethods
...
       # @example Basic Example
-      #   require 'hanami/validations'
+      #   require "hanami/validations"
       #
       #   class Signup
       #     include Hanami::Validations
       #
       #     validations do
-      #       required(:name).filled
+      #       required(:name).value(:string)
```
timriley: If you want this change to still represent the same behaviour as before, then it should be `filled(:string)` (not sure how important this is to you but I just wanted to note it!)

lib/hanami/validations/form.rb
```diff
@@ -12,57 +10,65 @@ module Validations
...
+    #   require "hanami/validations/form"
     #
     #   class Signup
     #     include Hanami::Validations::Form
     #
     #     validations do
-    #       required(:name).filled(:str?)
-    #       optional(:location).filled(:str?)
+    #       required(:name).value(:string)
+    #       optional(:location).value(:string)
```
timriley: Same deal here, these should be `filled(:string)` if you want the example to express the same behaviour as before

lib/hanami/validations.rb
jodosha: @timriley Thanks for letting me know 👍 https://github.com/hanami/validations/pull/180/commits/eef2c6da00000446c045fe94b951f27466a6911c

lib/hanami/validations/form.rb
jodosha: Fixed by https://github.com/hanami/validations/pull/180/commits/eef2c6da00000446c045fe94b951f27466a6911c

PR #196, opened 2020-04-01
Allow predicates and predicate together

pmatsinopoulos:
This is a fix to allow to use `predicate` and `predicates` together in the same class. Like this:

```
  class Validator
    include Hanami::Validations

    predicate :url?, message: 'must be an URL' do |current|
        current.start_with?('http')
    end

    predicates(
        Module.new do
            include Hanami::Validations::Predicates
            self.messages_path = "spec/support/fixtures/messages.yml"

            predicate(:email?) do |current|
              current.match(/@/)
            end
        end
    )

    validations do
        required(:foo) { url? | email? }
    end
  end
```
Without this fix, the above doesn't work. It does not find the `:url?` predicate, because it is not used. Only the module level predicates are used.


spec/unit/hanami/validations/predicates/schema/custom_spec.rb
```diff
@@ -120,6 +120,100 @@ def self.name
...
+        )
+
+        validations do
+          required(:foo) { url? | email? }
+        end
+      end
+    end
+
+    describe 'with valid input' do
+      let(:input) { { foo: %w[http://hanamirb.org foo@mailinator.com].sample } }
```
katafrakt, line 154: I don't think using `sample` in specs is a good idea. This introduces randomness to tests and when this feature breaks in the future (and for example start to only accept urls), it might be not immediately visible, because if url is randomly chosen, the test will pass. I think it's better to either have two attributes or create two specs (one for email, one for url).

pmatsinopoulos: Addressed here: b780b305dd6ce15fbed49e079890ec53661a9b81

spec/unit/hanami/validations/predicates/schema/custom_spec.rb
```diff
@@ -120,6 +120,108 @@ def self.name
...
+        end
+
+        validations do
+          required(:foo) { url? | email? }
+        end
+      end
+    end
+
+    describe 'with valid input' do
+      let(:input) { { foo: %w[http://hanamirb.org foo@mailinator.com].sample } }
```
katafrakt, line 209: One more place with `sample` in tests here.

pmatsinopoulos: Address here 7530bc38092d0d1cd631c2583d6285aceee88c61 and apologies

PragTob:
:wave: 

I was just opening an issue about this and before submitting it thought to check open PRs. Thanks @pmatsinopoulos and everyone else for your work! :green_heart: 

Can I help this forward anyhow? Would be great to have as this was mightily confusing for me :)

Cheers,
Tobi

pmatsinopoulos:
Hi @katafrakt ! Can I ask, please, what is the plan of merging this PR? Thank you very much in advance.

katafrakt:
@pmatsinopoulos no idea, I'm not on the core team.

pmatsinopoulos:
I don't know how to fix this PR. I understand that Ruby 2.3 should not be supported any longer. So, we will need to remove the Ruby 2.3 built from CI?

kaikuchn:
Hey @pmatsinopoulos, thanks for this PR I'll raise awareness in the Core team about it so that we get it reviewed & merged soon.
Regarding Ruby version support: I do believe we need to support Ruby 2.3+ for Hanami 1.x, however, it seems the problem with CI is not due to changes that you've made and I therefore believe you can disregard them.

jodosha:
Manually squashed and merged into https://github.com/hanami/validations/commit/a88405e5b750c2cb95d1e05f0a3f69a8f37bd29f because of conflicts of this PR.

@pmatsinopoulos Thanks for this fix, I preserved your commits during the manual process.

pmatsinopoulos:
> Manually squashed and merged into [a88405e](https://github.com/hanami/validations/commit/a88405e5b750c2cb95d1e05f0a3f69a8f37bd29f) because of conflicts of this PR.
> 
> @pmatsinopoulos Thanks for this fix, I preserved your commits during the manual process.

Great! Thank you very much!

PR #222, opened 2022-07-11
[v1.3] Pin `dry-container` at `<= 0.9`

cllns:
Similar to #221. The issue is that `dry-types` expects `dry-configurable` to be required, since it used to be required by `dry-container` but with 0.10, it no longer is.

This issue is causing `hanami new` to fail: https://github.com/hanami/hanami/issues/1182

### DIFF
--- a/hanami-validations.gemspec
+++ b/hanami-validations.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "dry-validation",   "~> 0.11", "< 0.12"
   spec.add_dependency "dry-logic",        "~> 0.4.2", "< 0.5"
   spec.add_dependency "dry-configurable", "<= 0.12"
+  spec.add_dependency "dry-container", "<= 0.9"
 
   spec.add_development_dependency "bundler", ">= 1.6", "< 3"
   spec.add_development_dependency "rake",    "~> 13"


salsysd:
so does this code commit fix the issue? trying hanami for the first! not best impressions thus. but it happens just curious if I can pull this

cllns:
@salsysd We're focusing our efforts on Hanami 2 (and will be releasing out first beta for it soon!). But we still want to support Hanami 1.3, of course. See this comment for a workaround: https://github.com/hanami/hanami/issues/1182#issuecomment-1181858526

@solnic Hanami 1.3 is stuck on dry-types 0.12, which does [use `Dry::Configurable`](https://github.com/dry-rb/dry-types/blob/release-0.12/lib/dry/types.rb#L25). It has access to that constant from `require 'dry-container'` (when dry-container is less than 0.10). This underlying issue could also be fixed by adding `require 'dry-configurable'` and releasing a new dry-types 0.12 gem, but I understand that that version is _long_ unsupported_ so it feels like something we can just workaround in Hanami. If you look at the stacktrace from https://github.com/hanami/hanami/issues/1182 you can see the affected versions. I'm able to reproduce it locally (and sticking with dry-container 0.9 is a valid workaround)

cllns:
It looks like, in practice, this will only ever get up to `dry-container` `0.8` due to it being the last `dry-container` that supports `dry-configurable` at `0.12`. Should I lower the version here to match that?

cllns:
(Clicked the wrong button, didn't mean to close this)

jodosha (approved):
@cllns Thanks Sean! 🙌

PR #226, opened 2022-11-18
Prepare README and API docs for 2.0

timriley:
Because this gem is currently only used as an internal support gem, with `Hanami::Action.params` being the primary public interface for action params validation, everything here is marked as `@api private`. This will allow us to adjust this gem (possibly stripping it back further) later during the 2.0 release cycle.


timriley:
[Prepare API docs for hanami-validations](https://trello.com/c/I9tReYzp/266-prepare-api-docs-for-hanami-validations)

jodosha (approved):

Issue #228, opened 2022-12-27
Missing CSRF token makes params invalid in test


kukicola:
Hello
In tests, CSRF token isn't set (and it's expected behavior) but calling `request.params.valid?` will return `false` because param is blank (even if my params validation rules don't include `_csrf_token` field):

```
> request.params.valid?
=> false
> request.params.errors
=> {:_csrf_token=>["must be filled"]}
```

I believe it's caused by:
https://github.com/hanami/controller/blob/main/lib/hanami/action/params.rb#L116
or
https://github.com/hanami/validations/blob/main/lib/hanami/validations/form.rb#L14

To fix it, I could remove input with CSRF token in views in test env or include middleware:

```
class RemoveCsrfParamMiddleware
  def initialize(app)
    @app = app
  end

  def call(env)
    return @app.call(env) if env["CONTENT_TYPE"]&.start_with?("multipart")

    params = Rack::Utils.parse_query(env["rack.input"].read, "&")
    params.delete("_csrf_token")
    env["rack.input"] = StringIO.new(Rack::Utils.build_query(params))
    @app.call(env)
  end
end
```

I think that in test env this default param validation should be removed

PR #229, opened 2023-02-08
Don't validate CSRF Token when Hanami env is `:test`

jodosha:
Fixes https://github.com/hanami/validations/issues/228


jodosha:
The manual test I did with the [demo app](https://github.com/kukicola/hanami-csrf-example) confirms that the fix works.
I need to make the specs to pass.

